Issue #201, opened 2015-10-21
Create proper unit tests for waca


Matthewrbowker:
The framework is in place - let's create proper unit tests for waca.

Matthewrbowker:
Moving it down the line a little bit, no way I'm going to get this done by June 1 (as it's June 9 xD)

Matthewrbowker:
I'm going to reclaim this issue, as I can begin working on this once #571 is merged.  

I'd like to call this issue resolved once we hit 80% coverage.  Is that OK?

Issue #184, opened 2015-05-24
Create coding style guide and stick to it


FunPika:
Following up on a comment @stwalkerster left on #182, we should really agree on a standard way of indenting lines. Once we do this, we should go through all of the code and make sure it uses that standard (Scrutinizer can probably do this automatically).

I personally would prefer to use [MediaWiki's tab style] (https://www.mediawiki.org/wiki/Manual:Coding_conventions#Tab_size), which is one hard tab per indentation level.

stwalkerster:
I think we're there in #232, so will close for now

PR #221, opened 2016-01-16
Fix recursive bug

cyberpower678:
Fix a recursive when OAuth if off.


stwalkerster:
Jenkins can't build this automatically.

cyberpower678:
What does that mean?

Matthewrbowker:
Don't worry about it, it's for #201 that I'm doing. 

Matthew Bowker

.gitignore
```diff
@@ -29,3 +29,4 @@ web.*.config
 waca.sln.DotSettings.user
 .vs/*
 .arc*
+password.php
```
stwalkerster: Be careful about ignoring .php files in a PHP project :grinning:

I'd be tempted to add this it `.git/info/exclude` since it's specific to your exact configuration

cyberpower678: I'm assuming your not going to add a file called password.php, if security through obscurity is your concern. :p

oauth/OAuthUtility.php
```diff
@@ -146,8 +146,8 @@ public function apiCall($apiParams, $accessToken, $accessSecret, $method = "GET"
 
 	public function getIdentity($accessToken, $accessSecret)
 	{
-
-		global $toolUserAgent;
+		
+        global $toolUserAgent;
```
stwalkerster: Why the whitespace change? #184 is strongly leaning tabs at the moment.

cyberpower678: IDK what happened.  I tried to fix it but it wouldn't fix.

.gitignore
FastLizard4: I think @stwalkerster is saying that you shouldn't add this file to .gitignore since you're the only person who needs it.  Instead, add it to `.git/info/exclude`, which is like .gitignore but only for you and never pushed to the rest of the repository.

Also, on the subject of whitespace, please remember to include a blank line at the end of files (with the exception of minified JS and CSS files).

includes/DataObjects/User.php
```diff
@@ -644,7 +644,7 @@ public function isAdmin()
     
 	public function isCheckuser()
 	{
-		return $this->checkuser == 1 || $this->oauthCanCheckUser();
+		return $this->checkuser == 1 || ($this->isOAuthLinked() && $this->oauthCanCheckUser());
```
FastLizard4: Looking at the definition of oauthCanCheckUser(), I don't see that this actually changes anything.  What recursion does this pull request fix?

oauth/OAuthUtility.php
FastLizard4: I'd suggest figuring out how to fix this before merging this pull request.

includes/DataObjects/User.php
stwalkerster: It's not a recursion fix at all, it's a fix for when the user is *not* linked to OAuth (and when that's not forced, such as development systems), it still does the OAuth call to get the groups and see if checkuser is one of them. I've been unable to reproduce it on my system (I use OAuth), but I spotted it while debugging. `isOAuthLinked()` returns a false if OAuth isn't configured for that user, thus this is simply short-circuiting the OAuth check.

oauth/OAuthUtility.php
FastLizard4: Actually, @cyberpower678, since you don't make any other changes in this file, you can also just remove it from the pull request so the whitespace changes aren't submitted.

includes/DataObjects/User.php
FastLizard4: That's what I figured, but the title of the pull request confused me.  Are there any other places we may want to short-circuit OAuth checks in this fashion?

stwalkerster: I thought I'd already caught them all, then this one showed up. Probably worth a hunt around.

FastLizard4: Maybe a better approach would be to add a isOAuthLinked() check in the actual methods themselves (e.g., oauthCanCheckUser()) instead of hunting around looking for calls to the methods and short-circuiting there?

cyberpower678: When it ran on my same it appeared to run in a recursive loop.  But I could be mistaken.  One thing was for certain.  It was running in an infinite loop.

PR #222, opened 2016-01-17
Check that an account is OAuth linked before querying permissions

FastLizard4:
Alternate fix for enwikipedia-acc/waca#221

### DIFF
--- a/includes/DataObjects/User.php
+++ b/includes/DataObjects/User.php
@@ -732,7 +732,10 @@ SQL
 
 		return false;
 	}
-    
+
+	/**
+	 * @return bool
+     */
 	public function isOAuthLinked()
 	{
 		if ($this->onwikiname === "##OAUTH##") {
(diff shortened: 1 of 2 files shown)


includes/DataObjects/User.php
```diff
@@ -822,8 +825,15 @@ public function oauthCanCreateAccount()
 		}
 	}
 
+	/**
+	 * @return bool
+	 */
 	public function oauthCanCheckUser()
```
stwalkerster: may be worth making this protected if it's only every called from here and CommunityUser.

FastLizard4: Hrm, that might be quite a rabit hole though.  There's probably a *lot* of stuff we can make more private in various places.

stwalkerster: True. Again this is OAuth code which I dislike. Meh.

:shipit:

FastLizard4: On the other hand, we might as well start somewhere.  Made oauthCanCheckUser protected.

FastLizard4:
YOLO

PR #227, opened 2016-01-17
Minor DB performance improvements

cyberpower678:
DB performance enhancements.  It's probably not much, but it should help.

### DIFF
--- a/includes/DataObjects/Request.php
+++ b/includes/DataObjects/Request.php
@@ -30,6 +30,9 @@ class Request extends DataObject
 	private $emailRequestsResolved = false;
 
 	private $blacklistCache = null;
+    
+	private static $commentsQuery = null;
+	private static $commentsResults = array();
 
 	/**
 	 * This function removes all old requests which are not yet email-confirmed
@@ -256,6 +259,11 @@ SQL
 		if ($this->hasCommentsResolved) {
 			return $this->hasComments;
 		}
+        
+		if (isset(self::$commentsResults[$this->id])) {
+		    $this->hasCommentsResolved = true;
+		    return $this->hasComments = self::$commentsResults[$this->id];
+		}
 
 		if ($this->comment != "") {
 			$this->hasComments = true;
@@ -263,12 +271,12 @@ SQL
 			return true;
 		}
 
-		$commentsQuery = $this->dbObject->prepare("SELECT COUNT(*) as num FROM comment where request = :id;");
-		$commentsQuery->bindValue(":id", $this->id);
+		if (is_null(self::$commentsQuery)) self::$commentsQuery = $this->dbObject->prepare("SELECT COUNT(*) as num FROM comment where request = :id;");
+		self::$commentsQuery->bindValue(":id", $this->id);
 
-		$commentsQuery->execute();
+		self::$commentsQuery->execute();
 
-		$this->hasComments = ($commentsQuery->fetchColumn() != 0);
+		self::$commentsResults[$this->id] = $this->hasComments = (self::$commentsQuery->fetchColumn() != 0);
 		$this->hasCommentsResolved = true;
 
 		return $this->hasComments;
--- a/includes/DataObject.php
+++ b/includes/DataObject.php
@@ -18,6 +18,10 @@ abstract class DataObject
 	public $isNew = true;
 
 	protected $dbObject;
+    
+	protected static $getByIdStatement = array();
+    
+	protected static $getByIdResults = array();
 
 	public function setDatabase(PdoDatabase $db)
 	{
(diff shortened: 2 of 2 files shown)


includes/DataObjects/Request.php
```diff
@@ -263,12 +271,12 @@ public function hasComments()
...
-		$commentsQuery = $this->dbObject->prepare("SELECT COUNT(*) as num FROM comment where request = :id;");
-		$commentsQuery->bindValue(":id", $this->id);
+		if (is_null(self::$commentsQuery)) self::$commentsQuery = $this->dbObject->prepare("SELECT COUNT(*) as num FROM comment where request = :id;");
+		self::$commentsQuery->bindValue(":id", $this->id);
 
-		$commentsQuery->execute();
+		self::$commentsQuery->execute();
 
-		$this->hasComments = ($commentsQuery->fetchColumn() != 0);
+		self::$commentsResults[$this->id] = $this->hasComments = (self::$commentsQuery->fetchColumn() != 0);
```
stwalkerster: One assignment per line makes it much more obvious what the code is doing.

includes/DataObjects/Request.php
```diff
@@ -263,12 +271,12 @@ public function hasComments()
 			return true;
 		}
 
-		$commentsQuery = $this->dbObject->prepare("SELECT COUNT(*) as num FROM comment where request = :id;");
-		$commentsQuery->bindValue(":id", $this->id);
+		if (is_null(self::$commentsQuery)) self::$commentsQuery = $this->dbObject->prepare("SELECT COUNT(*) as num FROM comment where request = :id;");
```
stwalkerster: Aside from the line length making this hard to review, it's also hard to see what's happening. Bytes are cheap - please use braces.

I'm also wondering if pushing this up to DataObject in a clean way would help alleviate the inevitable explosion of fields here - I'm thinking of a method where we can pass in a key, and it uses the current class name to look up in a cache. Again, I'd like other input on this before we go ahead, as it'll be a pretty big change. @FastLizard4 ?

includes/DataObjects/Request.php
```diff
@@ -256,6 +259,11 @@ public function hasComments()
 		if ($this->hasCommentsResolved) {
 			return $this->hasComments;
 		}
+        
+		if (isset(self::$commentsResults[$this->id])) {
+		    $this->hasCommentsResolved = true;
+		    return $this->hasComments = self::$commentsResults[$this->id];
+		}
```
stwalkerster: OK, so I'm slightly confused about this hunk of code.

If I'm right, this creates a global cache (`self::$commentsResults`) of whether the comments have been resolved, right, despite the local cache (`$this->hasComments`)?

I get the feeling that we might be better by default caching the returned object from the `getBy*()` functions so we only have to retrieve the object from the database once. At a global scale, I'm pretty sure there are some instances in the templates where we specifically do a database request direct from the template (urgh) along the lines of `{Foo::getById(4, gGetDb())->getBar()}`. Opinions @FastLizard4 ?

includes/DataObjects/Request.php
cyberpower678: What do you mean?  Did I do something wrong here?

includes/DataObjects/Request.php
cyberpower678: It's essentially the same commands but it's being sent to a static variable in the class.  This checks if the statement is already prepared.

includes/DataObjects/Request.php
cyberpower678: I may have interpreted the code incorrectly, but each request has it's own class correct?  The static variable preserves the information across all the classes, so if another duplicate class is opened it forgoes the querying and returns the information already cached.  It may be redundant perhaps, but it shouldn't impact performance and operation.

includes/DataObjects/Request.php
stwalkerster: It's much easier to read and hence understand and debug if you only put one assignment per line. Thus:

```php
$this->hasComments = (self::$commentsQuery->fetchColumn() != 0);
self::$commentsResults[$this->id] = $this->hasComments;
```

cyberpower678: Oh.

includes/DataObjects/Request.php
stwalkerster: Each request has it's own instance of this class, but we do nothing as far as I'm aware to prevent there being multiple instances of this class representing the same request. I can't think of any point (not checked) where we would actually want multiple objects representing the same database record, which is why I'm suggesting caching the entire object, rather than just specific fields. That way there's both memory and cpu time saved.

FastLizard4:
I haven't had a chance yet to review this PR in detail; however, at a cursory glance, I see several things in here that I don't like and need to examine further.  Until then, I am against merging this pull request.

Please also address the code style issues raised by @stwalkerster.

cyberpower678:
Like what?

includes/DataObjects/Request.php
FastLizard4: Yeah, I'm inclined to say cache the entire object, instead of specific fields.

includes/DataObjects/Request.php
FastLizard4: Pushing this up to DataObject is a big change, but it's the best way to do it.

FastLizard4:
Okay, now that I've finally had a chance to properly review this patch, I am against merging of this pull request.

For one, as noted by Simon, the formatting of your code in some parts makes it difficult to read.  Easily readable code is good code.

More importantly, though, I take issue with your usage of static fields.  Static fields are meant to be just that - stateless, constant values that can be used independent of the state of any objects of that class.  By using them for caching, you are making them both non-constant, and using them to store state information.  Put simply, it's an incorrect usage of static fields, and is contrary to good programming practices, as effectively, you've turned them into global variables.

Even if this code works, it's not a good solution to the problem of caching prepared statements.  As evidenced by the fact that we've had to shut down the tool for emergency maintenance, the last thing we need for ACC right now is more hard-to-read code that takes bad approaches to solving problems.  If anything, we should be rewriting the tool to be more properly object-oriented, not less.

I encourage you to investigate more "correct" approaches.  One approach I will suggest is writing a class to handle caching of objects returned from the database, instead of the prepared statements themselves (or even modifying one of our existing classes to do this) - this would obviate the need for caching the individual prepared statements.

For these reasons, I am closing this pull without merging.  If anything about my comments are unclear, I'd be happy to go through it with you once we've got the current situation with the tool handled.  (I imagine @stwalkerster and I will also start looking into caching of various things once the current situation is handled.)

Issue #215, opened 2016-01-04
"Deferred to admins" comment when deferring a request to Flagged Users


MJ94:
When I deferred request 160196 to Flagged Users today, the comment log marked it as "deferred to admins." This should say "deferred to flagged users," correct? Many flagged users are not tool administrators.

cyberpower678:
That log entry probably goes back to the time when accounts creators didn't exist as a right, so admins had to do the overriding of limits and spoofs.

Gesendet von Maximilian's iPhone. 
(Sent from Maximilian's iPhone.)

MJ94:
I'm referring to the tool comment log, not any log on Wikipedia.

cyberpower678:
Yea I meant that too.  Sorry. 

Gesendet von Maximilian's iPhone. 
(Sent from Maximilian's iPhone.)

FastLizard4:
So, the problem here is in the ACC configuration, actually:

```php
$availableRequestStates = array(
	// ...
	'Admin'=>array(
		'defertolog' => 'admins', // don't change or you'll break old logs
		'deferto' => 'flagged users',
		'header' => 'Flagged user needed',
		'api' => "admin",
		),
	// ...
);
```

It's the `'defertolog' => 'admins'` that's causing this problem.  Unfortunately, as noted, one can't change the defertolog option without breaking old logs.  So fixing this will probably involve doing some database modifications.

Paging @stwalkerster!

FastLizard4:
Looks like, at minimum, changing the 'defertolog' value will necessitate this query:

```sql
UPDATE `log`
SET `action`='Deferred to flagged users'
WHERE `action`='Deferred to admins';
```

FastLizard4:
Upon further review of the ACC code, I think changing the 'defertolog' value should actually be pretty safe - it should just be a matter of changing the value in config, then running the above database query.

FastLizard4:
So, this seemed to work fine in my local development environment.  Of course, that's a far cry from production... but at least that means this should be doable.  What do you think, @stwalkerster?

stwalkerster:
Change `config.inc.php` (tool config doesn't override that setting), and create a DB patch with @FastLizard4 's query and you're done. :grinning:

stwalkerster:
While you're at it, it might be an idea to change the array key of that entry too, and add another DB change:

```sql
UPDATE request
SET status = 'Flagged users'
WHERE status = 'Admin';
```

[...]

PR #235, opened 2016-01-31
When auto-deferring requests, send them to the new queue

FastLizard4:
Finishes fixing #215 

We should probably have a config option for this, but since @stwalkerster is currently on rewriting the config system, I'll leave that to him.  :P


stwalkerster:
:shipit:

PR #239, opened 2016-02-06
Close another XSS hole

stwalkerster:
Fixed in the same way as the other two you fixed :smile:


FastLizard4:
Aha.  Easy enough to fix, fortunately!

PR #251, opened 2016-03-11
Look up default action for close reason for created check

### LINKED ISSUE #250: Option to check if account is created on custom closes
Related to #247, but in a more specific use case.  Specifically with the "Created (COI risk)" closure on the live config, there is no check to ensure that the account has actually been created.  Maybe this is something we can configure on a per-email basis?

stwalkerster:
This should partially fix #250 for the legacy code. This will not fix the issue for custom closes, that's going to be a lot more challenging, but this should fix the initially reported problem.

### DIFF
--- a/acc.php
+++ b/acc.php
@@ -1189,9 +1189,13 @@ elseif ($action == "done" && $_GET['id'] != "") {
 	else {
 		$exists = false;
 	}
-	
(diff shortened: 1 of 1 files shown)


FastLizard4:
Tested to work as expected.  +2

PR #256, opened 2016-03-22
Create a couple tests and fix #126

Matthewrbowker:
This is a real quick pull request, including functionality for issue #126 and a few basic tests.

Issue #126, opened 2014-02-10
Offer buttons to Google email or visit ISP


Technical-13:
Should be able to Google the email address and if not the generic free hosts (Yahoo, Hotmail, Gmail, etc) should be able to navigate to the home page of the domain.

Matthewrbowker:
Coming when Pull request #232 lands.

PR #257, opened 2016-03-23
Include buttons to google email or visit email domain.  Fix for #126.

Matthewrbowker:
This is a revised version of pull request #256, with just the fix to #126 in.

### DIFF
--- a/templates/zoom-parts/username-section.tpl
+++ b/templates/zoom-parts/username-section.tpl
@@ -16,7 +16,8 @@
     <a id="UsernameGoogleSearch" class="btn btn-small" target="_blank" href="{$baseurl}/redir.php?tool=google&amp;data={$usernamerawunicode|escape:'url'}" onMouseUp="$('#UsernameGoogleSearch').addClass('btn-visited');">Google search</a>
     {if $showinfo}
       <a id="EmailVisit" class="btn btn-small" target="_blank" href="http://{$emailurl}" onMouseUp="$('#EmailVisit').addClass('btn-visited');">Visit email domain</a>
-      <a id="EmailGoogle" class="btn btn-small" target="_blank" href="{$baseurl}/redir.php?tool=google&amp;data={$emailurlencode|escape:'url'}" onMouseUp="$('#EmailGoogle').addClass('btn-visited');">Google email domain</a>
+      <a id="EmailGoogle" class="btn btn-small" target="_blank" href="{$baseurl}/redir.php?tool=google&amp;data={$request->getEmail()|escape:'url'}" onMouseUp="$('#EmailGoogle').addClass('btn-visited');">Google email</a>
+      <a id="EmailDomainGoogle" class="btn btn-small" target="_blank" href="{$baseurl}/redir.php?tool=google&amp;data={$emailurl|escape:'url'}" onMouseUp="$('#EmailDomainGoogle').addClass('btn-visited');">Google email domain</a>
     {/if}
 
   </div>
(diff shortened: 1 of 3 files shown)


zoompage.php
```diff
@@ -96,6 +96,10 @@ function zoomPage($id, $urlhash)
 	if ($hideinfo == false || $correcthash == true || User::getCurrent()->isAdmin() || User::getCurrent()->isCheckuser()) {
 		$smarty->assign("showinfo", true);
 	}
+
+	$emailDomain = explode("@", $request->getEmail())[1];
+	$smarty->assign("emailurl", $emailDomain);
+	$smarty->assign("emailurlencode", urlencode($emailDomain));
```
FastLizard4: this urlencode() call is unnecessary with `|escape:'url'` given in the Smarty template.

FastLizard4:
A comment in addition to the inline comment in the diff:

* #126 specifies that the option to Google should be for the entire email address, not just the email domain (maybe have buttons for both, but I'm inclined to think that Googling for the entire email address is going to be more useful)

stwalkerster:
I actually think there may be an xss vuln here, albeit very hard to exploit. Will need to check later tonight.

Matthewrbowker:
@FastLizard4 I have addressed your comments.  @stwalkerster I don't know what to say to the XSS vulnerability, but I would be concerned as the email is displayed further up the page too.

templates/zoom-parts/username-section.tpl
```diff
@@ -14,6 +14,12 @@
     <a id="UsernameUsernameList" class="btn btn-small" target="_blank" href="https://en.wikipedia.org/w/index.php?title=Special%3AListUsers&amp;username={$usernamerawunicode|escape:'url'}&amp;group=&amp;limit=1" onMouseUp="$('#UsernameUsernameList').addClass('btn-visited');">Username list</a>
     <a id="UsernameMainspaceSearch" class="btn btn-small" target="_blank" href="https://en.wikipedia.org/w/index.php?title=Special%3ASearch&amp;profile=advanced&amp;search={$usernamerawunicode|escape:'url'}&amp;fulltext=Search&amp;ns0=1&amp;redirs=1&amp;profile=advanced" onMouseUp="$('#UsernameMainspaceSearch').addClass('btn-visited');">Wikipedia mainspace search</a>
     <a id="UsernameGoogleSearch" class="btn btn-small" target="_blank" href="{$baseurl}/redir.php?tool=google&amp;data={$usernamerawunicode|escape:'url'}" onMouseUp="$('#UsernameGoogleSearch').addClass('btn-visited');">Google search</a>
+    {if $showinfo}
```
stwalkerster: I do wonder if this is the right place for this... it'd almost make sense for a new section. Maybe it's something for me to think about when doing the merge to newinternal

templates/zoom-parts/username-section.tpl
```diff
@@ -14,6 +14,12 @@
...
+      <a id="EmailVisit" class="btn btn-small" target="_blank" href="http://{$emailurl}" onMouseUp="$('#EmailVisit').addClass('btn-visited');">Visit email domain</a>
```
stwalkerster: So here's the possible XSS that I was thinking about - `$emailurl` is a substring of unsafe data, hence is probably also unsafe.

This is probably reasonably easily exploitable when email confirmation is disabled (ie: not production), but it may be exploitable with it enabled, depending on MTAs - I just don't know. It's probably safer to throw an `|escape` on this. This might warrant a check in code for a DNS record first though and only display if DNS resolves - @FastLizard4 thoughts?

zoompage.php
```diff
@@ -96,6 +96,9 @@ function zoomPage($id, $urlhash)
 	if ($hideinfo == false || $correcthash == true || User::getCurrent()->isAdmin() || User::getCurrent()->isCheckuser()) {
 		$smarty->assign("showinfo", true);
 	}
+
+	$emailDomain = explode("@", $request->getEmail())[1];
```
stwalkerster: `test"testy @ test"@example.com` is a valid email address IIRC. For that matter, I _believe_ `test@example.com(foo@bar)` is too.

Yes, email addresses are horrid. I don't know if we want to attack this specific scenario, but it's something to bear in mind.

stwalkerster:
Despite my comments, I'm reasonably happy with this - I'd feel much better if that `|escape` was there, but looks good to me. We probably ought to come to a decision about what we want to do about crazy but in-spec email addresses with the tool at some point too, but that's outside the scope of this PR.

templates/zoom-parts/username-section.tpl
FastLizard4: One approach you could take is use the redirect tool we already use for other Labs tools, and this has the benefit of also removing referrer data.  Pass the URL-encoded domain to the redirect tool, have the redirect tool check the domain, then if it's okay redirect the user.

zoompage.php
FastLizard4: Massive edge case :P

templates/zoom-parts/username-section.tpl
stwalkerster: That's a damn good point actually, this needs to use the redirect tool somehow, as otherwise we do leak data here. We probably want custom logic in there somewhere though, since at the moment it's quite dumb; a new mode flag switching it from "tool + data" to "untrusted domain" or something maybe?

zoompage.php
stwalkerster: Yeah, but it's the edge cases that introduce security holes. This I think is fine as it is, but the usage needs the |escape.

Matthewrbowker:
@stwalkerster @FastLizard4 I made a slight tweak to the redirection tool.  I'm not sure this is the best long-term solution - maybe we can make the redirect tool smart enough to check the domains as part of round 2?

Matthewrbowker:
Also done in stwalkerster/waca#4 for the newinternal branch.

Matthewrbowker:
I am closing this pull request, master is effectively feature frozen and this will land in newinternal.

PR #270, opened 2016-08-29
Add audit functionality

stwalkerster:
Adds auditing for failed logins, and both failed and successful password
resets. My testing shows that this seems to work as expected, though I can't test XFF functionality:
image

As per earlier emails, we could do with this change being approved from a privacy policy perspective. I don't expect it to have any issues since the data clear script is also changed to account for this, and this information is not visible.

We also *must* get this tested properly on the sandbox with the new grants, as they are a best-guess and not actually tested. We should test this by attempting any operation which I have revoked through the Web UI and seeing if that capability is exposed, as well as forcing the tool to use each privilege of each grant at least once.

I've added the table for ratelimiting, but I'm not looking at implementing that yet, I think it's more high-prio that we get audit functionality established. If this isn't reviewed/approved by the time I've got rate limiting done, then I'll add it, otherwise it can go in a separate PR.

+ Priority: High  - per emails,
+ Needs Discussion  - PrivPol check
+ live config  - grants need fixing on deployment

Targeting enwikipedia-acc/master for deployment *before* newinternal. I'll patch up newinternal to have the same added functionality in a little bit.

### DIFF
--- a/ClearOldData.php
+++ b/ClearOldData.php
@@ -19,7 +19,21 @@ $db->transactionally(function() use ($db)
 		":ip" => $cDataClearIp,
 		":mail" => $cDataClearEmail
 	));
-	
+
+	if (!$success) {
+		throw new TransactionException("Error in transaction: Could not clear data.");
+	}
+
+	$query = $db->prepare("DELETE FROM audit WHERE `timestamp` < DATE_SUB(curdate(), INTERVAL $dataclear_interval);");
+	$success = $query->execute();
+
+	if (!$success) {
+		throw new TransactionException("Error in transaction: Could not clear data.");
+	}
+
+	$query = $db->prepare("DELETE FROM ratelimit WHERE `timestamp` < DATE_SUB(curdate(), INTERVAL 1 DAY);");
+	$success = $query->execute();
+
 	if (!$success) {
 		throw new TransactionException("Error in transaction: Could not clear data.");
 	}
--- a/acc.php
+++ b/acc.php
@@ -287,6 +287,9 @@ elseif ($action == "forgotpw") {
 		$user = User::getByUsername($_POST['username'], gGetDb());
 
 		if ($user == false) {
+
+			auditForgottenPassword($_POST['username'], $_POST['email'], false);
+
 			BootstrapSkin::displayAlertBox(
 				"Could not find user with that username and email address!", 
 				"alert-error", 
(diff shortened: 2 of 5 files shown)


cyberpower678:
I think it would be acceptable to also allow the user to see parts of an audit log that affects them.  For example I could see the failed login attempts of my account.  Other high security sites routinely do this for the account owner's security.

stwalkerster:
@cyberpower678 Yeah, sure, but not for this PR. That can be something we add later on when newinternal is deployed. This is the intention of providing the functionality that can be monitored immediately, rather than providing any UI which would require much more security in place than this is intended to provide.

stwalkerster:
(By that I mean the security on this is "it's not accessible from the web UI, thus you must have db access to the firewalled DB to get at it")

FastLizard4:
Gonna wait until we get approval from DeltaQuad before I start code review/testing.

dqwiki:
There needs to be a useragent component to this as well. While I understand the need for privacy, if we are going to be preventing actual hijack attempts, then we are going to need that too.

To-do before testing/code review:
- [ ] Finish the ACC Applicant Privacy Policy (Assign: DeltaQuad)
- [ ] Finish the ACC Tool User Privacy Policy with provisions on failed login attempts (Assign: DeltaQuad)
- [ ] Create 2 new privacy policy pages (one for applicants and one for tool users) to place the privacy policies on.
- [ ] Include WMF Labs Registration Disclaimer on user account registration before submitting
- [ ] Include WMF Labs General Disclaimer 
- [ ] Require checkbox agreement of Privacy Policy on registration and application
- [ ] Create a "policy updated" agreement page which users will be redirected to upon login & an update to the privacy policy, requiring them to agree to the updated policy (UTRS Example can be provided)
- [ ] Create a field in the users table to indicate the users positive agreement to the policy
- [ ] Add links to the logged in navbar to the tool user PrivPol
- [ ] In some (undetermined?) way provide a link to the Applicant Privpol in an (undetermined?) easily accessible way
- [ ] Create a script which automatically removes the data in this table after a ______ amount of time (TBD time)

Before pushing live:
- [ ] Provide a non-legalized reader friendly report on how their data is stored (Assign: DeltaQuad)
- [ ] Tests to make sure no injection/improper access is gained to new table
- [ ] Testing to make sure script autoremoval works

After pushing live:
- [ ] Auditing to make sure autoremoval works.

PR #274, opened 2016-10-01
Add grants script

stwalkerster:
This script will automatically fix all of the relevant mysql grants on a way
we can be sure that the same grants are applied across different environments.

To run, grab yourself a privileged session on the database server, and run:

CALL SCHEMA_GRANTS_REFRESH("production", "'production'@'%'", "maintenance");

This should give a sane level of permissions which is currently needed by our
architecture. Future work elsewhere will allow us to pass in "web" to get a further
reduced set of permissions for the web UI, and a diferent set of permissions for
maintenance scripts.

As this is entirely in the database, it should be easy enough for us to include in the
standard deployment runbook. This is being added to the main repo so other
developers can use the same grants as in production, and so that it can be a developer
responsibility to think about what grants are actually needed, and not just deferring it
to the DBAs.

### DIFF
--- a/sql/grants.sql
+++ b/sql/grants.sql
@@ -16,7 +16,7 @@ CREATE PROCEDURE SCHEMA_GRANTS_REFRESH(
     -- against privmode
     -- --------------------------------------------------------------------------------------------------
     DECLARE cur_select CURSOR FOR
-        SELECT 'antispoofcache' AS tbl UNION
+        SELECT 'antispoofcache' UNION
         SELECT 'ban' UNION
         SELECT 'comment' UNION
         SELECT 'emailtemplate' UNION
@@ -31,7 +31,7 @@ CREATE PROCEDURE SCHEMA_GRANTS_REFRESH(
         SELECT 'xfftrustcache';
 
     DECLARE cur_update CURSOR FOR
-        SELECT 'ban' AS tbl UNION
+        SELECT 'ban' UNION
         SELECT 'comment' UNION
         SELECT 'emailtemplate' UNION
         SELECT 'geolocation' UNION
@@ -43,7 +43,7 @@ CREATE PROCEDURE SCHEMA_GRANTS_REFRESH(
         SELECT 'welcometemplate';
 
     DECLARE cur_insert CURSOR FOR
-        SELECT 'antispoofcache' AS tbl UNION
+        SELECT 'antispoofcache' UNION
         SELECT 'applicationlog' UNION
         SELECT 'ban' UNION
         SELECT 'comment' UNION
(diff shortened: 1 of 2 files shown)


sql/grants.sql
```diff
@@ -0,0 +1,181 @@
...
+        IF cursor_done THEN
+            LEAVE revtable;
+        END IF;
+        SET @sqlText = vsql;
+        PREPARE statement FROM @sqlText;
+        EXECUTE statement;
+        DEALLOCATE PREPARE statement;
+    UNTIL cursor_done END REPEAT revtable;
+    CLOSE cur_revtable;
+
```
FastLizard4: SQL syntax literally causes me physical pain sometimes

sql/grants.sql
```diff
@@ -0,0 +1,181 @@
...
+        IF cursor_done THEN
+            LEAVE ldelete;
+        END IF;
+        SET @sqlText = CONCAT('GRANT DELETE ON ', schemaname, '.', tableName, ' TO ', targetuser);
+        PREPARE statement FROM @sqlText;
+        EXECUTE statement;
+        DEALLOCATE PREPARE statement;
+    UNTIL cursor_done END REPEAT ldelete;
+    CLOSE cur_delete;
+
```
FastLizard4: Ugh.  This feels so DRY to me, but I don't see any reasonable way to avoid having to repeat basically the same loop four times.  :|

sql/grants.sql
```diff
@@ -0,0 +1,181 @@
...
+    -- All done!
+END;;
+DELIMITER ';'
```
FastLizard4: No newline at end of file &lt;/pedant&gt;

Actually, is this something we stylistically care about for ACC's code?

sql/grants.sql
```diff
@@ -0,0 +1,181 @@
...
+    DECLARE tableName VARCHAR(64);
+    DECLARE vsql VARCHAR(4000);
+
+    -- --------------------------------------------------------------------------------------------------
+    -- PERMISSIONS
+    -- Add tables to these queries to assign grants as needed. Limit the rows by adding where clauses
+    -- against privmode
+    -- --------------------------------------------------------------------------------------------------
+    DECLARE cur_select CURSOR FOR
+        SELECT 'antispoofcache' AS tbl UNION
```
FastLizard4: Is the `AS tbl` necessary for these cursors?  It seems odd to include it in the cursors for SELECT, UPDATE, and INSERT but not the DELETE cursor.

sql/grants.sql
stwalkerster: The solution would be a second SP, but I don't want to go down that route.

sql/grants.sql
stwalkerster: Actually, most of that is just MySQL. PL/SQL is so much nicer.

Like, actually, not kidding:

```sql
CREATE PROCEDURE foo ( 
    snork IN VARCHAR2
  , blurble IN NUMBER 
)
IS
  CURSOR bar IS SELECT 'REVOKE ALL....' AS vsql FROM baz;

  rec bar%ROWTYPE;
BEGIN
  
  FOR rec IN bar LOOP
    EXECUTE IMMEDIATE rec.vsql;
  END LOOP;
  
END;
```

sql/grants.sql
stwalkerster: No, I was thinking I could use rowtypes and reference the column by name, but that's too sensible for MySQL. See my comment below regarding the rowtypes in PL/SQL.

sql/grants.sql
FastLizard4: > PL/SQL is so much nicer.

Now there's something you don't get to type every day.  :P

sql/grants.sql
FastLizard4: Yeah, which is fair

FastLizard4:
@stwalkerster Now that we have the new "squash and merge" and "rebase and merge" options, do you think we should use one of those instead of the old pull merging behavior?

stwalkerster:
OK, for smaller changes like this I think probably want to squash. Changes large enough to need merges back from master periodically / larger features / etc, let's merge. Use judgement to decide which is better in a given scenario.

I'll merge this in over the next few days when I'm happy that I'm gonna be around to un-break everything when this goes horribly wrong because I've forgotten something.

dqwiki:
@stwalkerster Poke this dead PR

stwalkerster:
This isn't happening in master.

PR #277, opened 2016-12-05
Miser mode enhancements

stwalkerster:
Couple of basic improvements to miser mode considering the current state of play in production

* Change the colour of the badge in miser mode
* Add list all button for those who are insane enough to click it. See commit message for more info.

### DIFF
--- a/acc.php
+++ b/acc.php
@@ -2035,6 +2035,43 @@ elseif ($action == "oauthattach") {
 		}
 	});
 }
+elseif ($action == "listall") {
+    global $availableRequestStates, $enableEmailConfirm;
+
+    $database = gGetDb();
+
+    if ($enableEmailConfirm == 1) {
+        $query = "SELECT * FROM request WHERE status = :type AND emailconfirm = 'Confirmed';";
+    }
+    else {
+        $query = "SELECT * FROM request WHERE status = :type;";
+    }
+
+    $statement = $database->prepare($query);
+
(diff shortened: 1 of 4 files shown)


stwalkerster:
Testing: https://phabricator.stwalkerster.co.uk/M3

acc.php
```diff
@@ -2035,6 +2035,43 @@
 		}
 	});
 }
```
FastLizard4: Minor code style: elseif should be on the same line as the closing brace of the previous block

acc.php
```diff
@@ -2035,6 +2035,43 @@
...
+    if ($enableEmailConfirm == 1) {
+        $query = "SELECT * FROM request WHERE status = :type AND emailconfirm = 'Confirmed';";
+    }
+    else {
+        $query = "SELECT * FROM request WHERE status = :type;";
+    }
+
+    $statement = $database->prepare($query);
+
+    if (isset($_GET['status']) && isset($availableRequestStates[$_GET['status']])) {
```
FastLizard4: Maybe it makes sense to move this conditional up to just after the `global $availableRequestStates`?  That way we don't have to spend time preparing the database statement unless we have everything we need.

acc.php
stwalkerster: All style fixes applied in newinternal branch. 😉

FastLizard4: fien

FastLizard4:
good enough™

PR #285, opened 2017-01-29
Some minor bugfixes for newinternal

stwalkerster:
* Use constants instead of magic strings for user status
* Add a provision to allow extra CSS and JS resources to be loaded on a page-by-page basis
* Fix unhandled exception in IndentificationVerifier
* Refactor and expand capabilities of SearchHelper

PR #286, opened 2017-01-29
Add Role-Based Access Control to the tool

stwalkerster:
Depends on #285 

Adds a much more granular permissions system in preparation for event management


stwalkerster:
@dqwiki & @FastLizard4 - I'm pre-emptively giving a rundown of these changes so hopefully this can be understood fairly quickly by the two of you when looking at the code.


This PR completely revamps the security architecture of the tool. Now, instead of things being "are you an admin?", they are built on a group permissions model. Every action on every internal page which is defined in the active router implicitly creates a new permission. These permissions must be assigned to a role in `RoleConfiguration`, and users can be assigned roles from User Management

Because the grants for access to specific pages are now completely separated from individual pages, there is minimal risk of touching security-specific code when a minor fix needs to be made. All code which determines which user is able to perform an action is centralised and well-tested, rather than being re-implemented in many places.

## Refresher on application architecture ##
The application is now essentially a task runner, where every web UI request, API request, or console-launched script triggers the entire application to perform a job, known as a "Task". These tasks come in three forms - API requests, page loads, or console scripts. Page loads are further split into two - public interface pages and internal interface pages.

This structure is represented in code as well - `TaskBase` is the parent class for all types of requests. Inheriting from that is `PageBase`, `ApiPageBase`, and `ConsoleTaskBase`; inheriting from `PageBase` is `PublicInterfacePageBase` and `InternalPageBase`.

These changes are specifically angled towards `InternalPageBase`, as none of the other types of request need access control. 

## Role definitions ##

In the [`RoleConfiguration`](https://github.com/enwikipedia-acc/waca/blob/rbac/includes/Security/RoleConfiguration.php) class, there is an (enormous) multi-level array of data which describes the permissions model for the entire application. The top level is a `role => array()` map, which defines which roles exist containing what grants. The next level down is a `page => array()` map, defining which pages are available, and finally an `action => allow/deny` map. Page names are directly taken from the class which implements them, and action names are defined by the name given to the method which implements that action (and is also defined in the request router). For example, if you wanted to grant a role "testyrole" the ability to change their preferences and password, you would add the following:
```php
$roleConfig = array(
    'testyrole' => array(
        PagePreferences::class => array(
            self::MAIN       => self::ACCESS_ALLOW,
            'changePassword' => self::ACCESS_ALLOW,
        ),
    )
);
```

There are also special settings in the roles which allow a role to be hidden by default (allowing role grants to be grouped and assigned atomically to different roles), configuration to allow specific roles to be edited by specific groups, and child roles from which the current role inherits all grants in addition to those it has been explicitly granted. For example, a tool root role which has all the rights of a user in the "admin" and "checkuser" roles, but is only grantable and revokable by other tool roots would have the following configuration:
```php
$roleConfig = array(
    'toolRoot'         => array(
        '_description' => 'A user with shell access to the servers running the tool',
        '_editableBy' => array('toolRoot'),
        '_childRoles'             => array(
            'admin', 'checkuser',
        ),
    ),
);
```

### Special notes on roles ###

On sign-up, a default role is granted by the signup page. This allows us to have different signup pages for different user classes, for example event coordinators who should have a much more limited view of the application and the data in it. This functionality is not yet used, but was implemented in anticipation of event coordination extensions.

There are two "special" roles - `public` and `loggedIn` - which are granted to absolutely everyone, and logged-in users only. This represents a shift in the way we treat user accounts - everyone can log in, even if your account is brand new. This means you can get into your preferences and update your password etc if you wish, and you can see menu items for various bits of public data such as some of the statistics pages, as well as connect to OAuth. It also means the "Meta" menu is visible to everyone, even logged-out users.

We also have the concept of "inactive" roles - roles which are granted to a user but are not "activated" because the user is suspended or not identified. This takes away the (technical) requirement to ensure everyone is identified before approving them, as they simply will not be able to access the protected roles if they are not identified, and thus won't have access to see anything. We do include sensible error messages to inform the user of this.

If additional checks are required which cannot be determined by a single page, then additional checks can be run by calling into the `SecurityManager`. This is frequently used to hide UI elements which are meaningless to the user, such as hiding user management from non-tool-admin users. Additionally, this is used by some additional checks on what data is visible - done with the special "RequestData" group of grants.

There is a way for a page to explicitly declare it is exempt from the access control system - this is used for pages like the login page, the 403 page, and the 404 page. It exists because these are things which should never be restricted by roles, as access to these pages is technically required. Because this is an explicit opt-out by overriding the `isProtectedPage()` method, it's obvious which pages are opted out by just finding all subclasses which override that method.

## Change overview ##

The changes are huge because of the number of places function calls such as `isAdmin()`, `isUser()`, and `isCheckuser()` were called. For the most part, calls to these functions have been removed and delegated to the managing page. The user state for "User" and "Admin" has been combined into "Active"; which also has the side effect that users promoted to tool admin do not lose this status when their account is suspended. Likewise, the column defining a user as a Checkuser has also been removed, as this has been converted into a role. A lot of the "get all admins"-type code from the User DataObject class has also been removed - partly because it's no longer relevant and partly because it doesn't belong in a class responsible for managing a single user - SearchHelpers are the replacement for these. 

`InternalPageBase` contains the main security barrier to running pages and testing access to actions.

`includes/ConsoleTasks/MigrateToRoles.php` is a one-shot migration script in place of the normal schema upgrade script. This script converts an existing user table into a role-based table, whereas the actual patch25 migration script assumes an empty database. We will run this console task in place of the patch25 schema upgrade script as part of the installation.

### Other notes ###
Some changes to templates include adding the `updateversion` field for optimistic locking. This is a bugfix, as this field should have already been present.

We also extend the searchability of logs as well taking advantage of the improvements to the SearchHelpers.

PR #420, opened 2017-03-12
Performance: cache the checkuser status of a user

stwalkerster:
Analysing some XHProf data has revealed that the `User::isCheckuser` function is exceedingly slow.

image

The old implementation of `return $this->checkuser == 1 || $this->oauthCanCheckUser()` means that the OAuth identity ticket will be checked on every call of this method, which involves an unserialize, date cache check, array search, and a possible remote fetch. 

image

Needless to say, this is something which we don't care if it changes within a request, so caching this is appropriate.

This, unfortunately, isn't really possible for me to test locally.

### DIFF
--- a/includes/DataObjects/User.php
+++ b/includes/DataObjects/User.php
@@ -30,6 +30,8 @@ class User extends DataObject
 
 	private $identityCache = null;
 
+	private $isCheckuserCache = null;
+
 	/**
 	 * Summary of getCurrent
 	 * @param PdoDatabase $database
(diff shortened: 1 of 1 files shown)


includes/DataObjects/User.php
```diff
@@ -652,7 +654,11 @@ public function isAdmin()
     
 	public function isCheckuser()
 	{
-		return $this->checkuser == 1 || $this->oauthCanCheckUser();
+	    if($this->isCheckuserCache === null) {
```
FastLizard4: Miiiiinor style note: Space after the `if`.  Not really important though.  Fix it if you want, otherwise go ahead and merge.  Also mix of spaces and tabs, but *shrug*.

PR #422, opened 2017-03-12
Make the webchat link in the header be HTTPS

issyl0:
- Freenode serves webchat over HTTPS, so I think we should take advantage of that.
- My editor removed trailing whitespace in this file. Let me know if you want me to revert those lines.


stwalkerster:
@FastLizard4 merged into master, deployed (fixing my tag screwup of yesterday too). Also merged into newinternal-bugfixing for newinternal. 

Cheers @issyl0 ! 😄

Issue #423, opened 2017-03-15
Change Limit from 1 to 25 for "Username List"


Dane2007:
Hello,

Could we update the limit from 1 to 25 for the username list button?

**Current Setup**
https://en.wikipedia.org/w/index.php?title=Special%3AListUsers&username=Dane&group=&limit=1

**New Setup**
https://en.wikipedia.org/w/index.php?title=Special%3AListUsers&username=Dane&group=&limit=25

Thanks,

Dane

Dane2007:
Issue committed and merged.

PR #424, opened 2017-03-29
Update username-section.tpl

Dane2007:
Change Limit from 1 to 25 for Username List button per Issue #423


stwalkerster:
(bypassed travis check due to travis being down)

PR #429, opened 2017-05-26
Unlock IP Search for users

Dane2007:
Now identified to Wikimedia Foundation.  This should be unlocked for certain research purposes.

### DIFF
--- a/search.php
+++ b/search.php
@@ -100,14 +100,6 @@ if (isset($_GET['term']) && isset($_GET['type'])) {
 		$smarty->display("search/searchresult.tpl");
 	}
 	elseif ($_GET['type'] == 'IP') {
-		// move this to here, so non-admins can perform searches, but not on IP addresses or emails
-		if (!User::getCurrent()->isAdmin() && !User::getCurrent()->isCheckuser()) {
-			// Displays both the error message and the footer of the interface.
-			BootstrapSkin::displayAlertBox("IP address search is only available to tool admins and checkusers.", "alert-error", "Access Denied");
-			$smarty->display("search/searchform.tpl");
-			BootstrapSkin::displayInternalFooter();
-			die();
-		}
         
 		$qterm = '%' . $term . '%';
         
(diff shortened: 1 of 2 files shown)


templates/search/searchform.tpl
```diff
@@ -11,11 +11,11 @@
       <select name="type">
         <option value="Request">... requested username</option>
         <option value="email">... email address</option>
+				<option value="IP">... IP address</option>
```
stwalkerster: You might want to fix the indentation here

templates/search/searchform.tpl
```diff
@@ -11,11 +11,11 @@
...
         {if $currentUser->isAdmin() || $currentUser->isCheckuser()}
```
stwalkerster: ... and while you're editing this file, let's remove the useless CIDR options and the duplicate IP option if you're a tool admin...

... which leaves the entire if block empty, and thus also unnecessary.

templates/search/searchform.tpl
```diff
@@ -11,12 +11,9 @@
...
         <option value="Request">... requested username</option>
         <option value="email">... email address</option>
+	<option value="IP">... IP address</option>
         {if $currentUser->isAdmin() || $currentUser->isCheckuser()}
         <option value="IP">... IP address</option>
-        <option value="CIDR" disabled="disabled">... CIDR Range</option>{* not implemented yet! *}
-        {else}
-        <option value="IP" disabled="disabled">... IP address</option>
-        <option value="CIDR" disabled="disabled">... CIDR Range</option>
         {/if}
```
FastLizard4: Lines 15 through 17 are now completely redundant and should be removed entirely; as is, the code will show the IP Address option to admins and checkusers twice.

search.php
```diff
@@ -100,14 +100,6 @@
...
 	}
 	elseif ($_GET['type'] == 'IP') {
-		// move this to here, so non-admins can perform searches, but not on IP addresses or emails
-		if (!User::getCurrent()->isAdmin() && !User::getCurrent()->isCheckuser()) {
-			// Displays both the error message and the footer of the interface.
-			BootstrapSkin::displayAlertBox("IP address search is only available to tool admins and checkusers.", "alert-error", "Access Denied");
-			$smarty->display("search/searchform.tpl");
-			BootstrapSkin::displayInternalFooter();
-			die();
-		}
```
FastLizard4: Code style pedant: Remove this blank line

PR #432, opened 2017-06-03
Geolocation pre-caching

### LINKED ISSUE #430: IpLocationProvider lag
IpLocationProvider's getResult method performs a web request to retrieve the IP location from a remote webservice. Occasionally, this appears to fall over, leading to page load times of between 30-90 seconds. 

We shouldn't be doing this on page load, unless absolutely necessary. We should probably push this to a background maintenance job via a probably hacky cronjob in master. We should also think about newinternal for this.

Alternatives include doing post-webrequest processing at request creation, or simply setting a timeout. I personally feel a short (5s?) timeout on pageload is necessary, in case the background processing isn't complete by the time the data is needed, we can still request the data at runtime. 95% of the time we should aim to have this data before it is needed.

```
IpLocationProvider::getResult       92,873,562 µs
```

stwalkerster:
Adds a new maintenance script which looks at all IP addresses in the
database and fetches the geolocation data for them all.

This will need to be scheduled via cron to have the desired effect, and
should be considered a quick-fix, not a long-term solution

Fixes #430 for master

### DIFF
--- a/PrecacheGeolocation.php
+++ b/PrecacheGeolocation.php
@@ -16,25 +16,31 @@ $database = gGetDb();
 
 $locationProvider = new IpLocationProvider($database, $locationProviderApiKey);
 
-$done = false;
-while (!$done) {
+while (true) {
     echo "Beginning txn\n";
     $database->beginTransaction();
 
     try {
+        // fetch a bunch of un-geolocated IPs from the database.
+        // note we have to parse the forwardedip field in the database so we can test against the geolocation table.
+        // This guarantees we get ten unlocated IPs back, unless there actually aren't 10 available.
+        // Alternatives include downloading a small set of forwarded IPs, splitting it in PHP, constructing an IN()
+        // clause dynamically, sending that back to the database to check if there are geolocation entries, then repeating
+        // until we have 10 to process - and the fact that we'd have to potentially retrieve all IPs from the database
+        // before we find any at all. This way keeps all of that legwork in the database, at the cost of a more complex
+        // query.
         $statement = $database->query(<<<SQL
             SELECT p.prox
             FROM (
               SELECT trim(substring_index(substring_index(r.forwardedip, ',', n.n), ',', -1)) prox
               FROM request r
-                INNER JOIN (SELECT 1 n
-                            UNION ALL
-                            SELECT 2
-                            UNION ALL SELECT 3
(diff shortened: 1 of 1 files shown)


PrecacheGeolocation.php
```diff
@@ -0,0 +1,78 @@
...
+            WHERE NOT EXISTS (SELECT 1 FROM geolocation g WHERE g.address = p.prox)
+            LIMIT 10;
+SQL
+        );
+
+        $missingIps = $statement->fetchAll(PDO::FETCH_COLUMN);
+
+        $count = count($missingIps);
+        if ($count === 0) {
+            $done = true;
```
FastLizard4: Seems like this $done is unnecessary, just use a while(true) since we're breaking out here anyway

PrecacheGeolocation.php
```diff
@@ -0,0 +1,78 @@
...
+    try {
+        $statement = $database->query(<<<SQL
+            SELECT p.prox
+            FROM (
+              SELECT trim(substring_index(substring_index(r.forwardedip, ',', n.n), ',', -1)) prox
+              FROM request r
+                INNER JOIN (SELECT 1 n
+                            UNION ALL
+                            SELECT 2
+                            UNION ALL SELECT 3
```
FastLizard4: Style suggestion: This SELECT should be on its own line, like the others

PrecacheGeolocation.php
```diff
@@ -0,0 +1,78 @@
...
+$done = false;
+while (!$done) {
+    echo "Beginning txn\n";
+    $database->beginTransaction();
+
+    try {
+        $statement = $database->query(<<<SQL
+            SELECT p.prox
+            FROM (
+              SELECT trim(substring_index(substring_index(r.forwardedip, ',', n.n), ',', -1)) prox
```
FastLizard4: Okay, I've been staring at this query and I'm still not sure what it's doing.  Since there are other requested changes here too, I'm going to need you to run me through it again.  :P

stwalkerster: **TL;DR:** It's splitting the `forwardedip` field from "a, b, c" to three rows: one each for `a`, `b`, and `c` using fancy substrings joined against a static list of numbers for distances.

Longer demo answer: https://phabricator.stwalkerster.co.uk/P137

PR #433, opened 2017-06-03
Switch username and IP sections

DatGuy1:
Issue #427

### DIFF
--- a/templates/view-request/main.tpl
+++ b/templates/view-request/main.tpl
@@ -25,9 +25,9 @@
         </div>
     </div><!--/row-->
 
-    {block name="ipSection"}{/block}
-
     {include file="view-request/username-section.tpl"}
+    
+    {block name="ipSection"}{/block}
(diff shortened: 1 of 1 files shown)


FastLizard4:
Why is travis failing this?

FastLizard4:
Also, I think this change needs some discussion from the tool users about whether or not this is a good thing; see also https://xkcd.com/1172/

stwalkerster:
From a code PoV, I think this is fine, but I want to check the UI for oddities, hence the delay in getting my dev env back online. I'm fairly sure there was consensus for this from IRC when I created the task.

As to Travis - probably something unrelated in newinternal, that's probably been fixed in newinternal-bugfixing. I'll look properly later.

stwalkerster:
I'll also target this for either newinternal-bugfixing, or possibly even something later like rbac, but I'll look at that from my dev env.

Issue #261, opened 2016-04-08
"Disable automatic welcoming" should always be available


Matthewrbowker:
Right now, it's unclear when a user has welcoming disabled.  It would be helpful to always display the option to "disable automatic welcoming" even if the user has selected it.

stwalkerster:
Simple case of removing the if statement from the `<tfoot>` of `templates/welcome-template/list.tpl`

@Matthewrbowker - do you have code for this already, or is this something a newbie could pick up? It's been assigned to you for over a year :)

Matthewrbowker:
@stwalkerster Erm... I honestly don't remember what this task is - I was holding off until newinternal was done but...

PR #446, opened 2017-06-15
Always show disable automatic welcoming button

DatGuy1:
Apparently I overcomplicated it. It would be good to just have a checkbox to tell if it is enabled and uncheck it to disable the feature. #261

### DIFF
--- a/resources/styles.css
+++ b/resources/styles.css
@@ -102,15 +102,16 @@ h5.zoom-button-header {
     filter: progid:DXImageTransform.Microsoft.gradient(startColorstr='#ff0088cc', endColorstr='#ff0077b3', GradientType=0);
 }
 
-td.success{
+
+td.success, th.success, .table tfoot tr.success > td, .table tfoot tr.success > th {
     background-color: #dff0d8;
 }
-td.error{
+td.error, th.error, .table tfoot tr.error > td, .table tfoot tr.error > th {
     background-color: #f2dede;
 }
-td.warning{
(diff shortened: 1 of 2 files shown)


stwalkerster:
OK, so my quick testing of this has revealed a flaw in just removing the `if` statement. When a welcome template is selected, note the green styling (from `class="success"`) of the row, and the change in visual style of the selected button (from the `if` on line 106) of `templates/welcome-template/list.tpl`:

image

However, when I choose the disable option, I get this:

image

We basically need to copy the styling applied to the selected button with an appropriate condition (`== null`), and make sure it's showing the right thing. I think we can ignore the green colouring for now, as that doesn't make much UX sense on when disabled.

DatGuy1:
I don't think I quite understand. So, if automatic welcoming is disabled, the row should have a green 'overlay' and have selected instead of select?

stwalkerster:
@DatGuy1 yeah, something like that. Basically, at the moment, when there's no template selected, the row is hidden, so there's no need for it to change to show it's selected like all the other ones do. With your change as it is at the moment, there's nothing to show that the "disable welcoming" row is the currently selected one.

As all of the other rows have a green background, and change to a disabled "selected" button when it's the one selected, this one should do the same thing for consistency.

stwalkerster:
@DatGuy1 this is what I meant:

image

Turns out that a quick attempt at implementing it was more problematic than I first thought, since the Bootstrap styles didn't apply to the `<tfoot>` element in HTML, so I also extended Bootstrap to add that styling.

PR #448, opened 2017-06-19
Add enhanced log search capabilities

stwalkerster:
Adds the capability to search logs by object type and object id.


stwalkerster:
Marked as blocking go-live because this also fixes a bug in the log page when you scroll back far enough to hit bans which were deleted from the database in the Ancient Times(tm)

Issue #451, opened 2017-06-28
Request::cleanExpiredUnconfirmedRequests occasionally takes excessive time


stwalkerster:
No idea what the root cause of this is, but we probably ought to move this task to the dataclear maintenance script anyway, rather than running what is essentially a maintenance task on page-load, especially public-side page load.

This specific request was email-confirmed four times due to the excessive page-load time.

![image](https://user-images.githubusercontent.com/722710/27637534-fa6012e6-5c07-11e7-9c43-4690c893b1cd.png)

I'm gonna mark this as easy, as it should just be a case of actioning the TODO near the top of `index.php` to move the call to `Request::cleanExpiredUnconfirmedRequests` to the `ClearOldData.php`.

I'll take a look and see if I can find a root cause later on, but I doubt we have the necessary information held to figure out what happened.

PR #452, opened 2017-06-28
Move Request::cleanExpiredUnconfirmedRequests to maintenance script

DatGuy1:
#451

### DIFF
--- a/ClearOldData.php
+++ b/ClearOldData.php
@@ -12,7 +12,7 @@ $db = gGetDb( );
 
 $db->transactionally(function() use ($db)
 {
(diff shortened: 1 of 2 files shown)


ClearOldData.php
```diff
@@ -20,6 +20,10 @@
 		":mail" => $cDataClearEmail
 	));
 	
+	if ($enableEmailConfirm == 1) {
```
FastLizard4: This code will cause a syntax error because you didn't add `$enableEmailConfirm` to the `global` directive on line 15.  Please remember to test your code to avoid issues like this.

FastLizard4: Correction, not a syntax error, but an undefined variable error.

PR #465, opened 2017-08-28
Perform sanity checks on XFF data of requests

FastLizard4:
Validate that the data we're being given in the X-Forwarded-For headers is sane - specifically, only keep items in the chain that are actually valid IP addresses, and silently discard anything that isn't.  Hopefully this will resolve some deadlock errors we've been seeing on ACC recently, as the presence of such an invalid XFF entry in a recent request was tripping up the geolocation pre-cacher and thus causing all manner of undesired behavior.

Remember kids, don't trust user input!

## Testing
X-Forwarded-For header used: `96.126.96.9, 23.21.22.11, unknown, bad, 392.123.111.999, 10103.3817894.41919.3781661, 8.8.8.8`

Without this pull request:
image

With this pull request:
image


FastLizard4:
@dqwiki Can you provide details about what issues this causes?  I've thoroughly tested this code, and the current implementation has the spaces, and the spaces are called for in the standard at https://tools.ietf.org/html/rfc7239#section-5.2

stwalkerster:
@dqwiki We split the string on `,`, then trim (L263) the spaces off the entries, then merge it back together with a `', '`

PR #466, opened 2017-09-28
Geoloc lock

stwalkerster:
Related to #465, this should also nerf the geolocation deadlocks.

### To recreate
Firstly, make sure you have an invalid XFF entry so the script loops due to a different bug.

Then:
```sql
SET SESSION sql_mode = 'ansi'; -- boo mysql
SET TRANSACTION ISOLATION LEVEL SERIALIZABLE;
START TRANSACTION ;
SELECT '\'' || forwardedip || '\',' FROM request WHERE forwardedip IS NOT NULL AND forwardedip NOT LIKE '%,%' LIMIT 10 FOR UPDATE;
```
Copy-pasta the resultset from the select into this (my version of ~~mysql~~ maria doesn't support LIMIT within the subquery(!)):
```sql
DELETE FROM geolocation WHERE address IN (/* copy-pasta */);
```

Launch multiple (I used three) copies of the `PrecacheGeolocation.php` script. They should all hang very rapidly as they wait for locks.

Finally:
```sql
COMMIT;
```

Sit back, relax, and watch everything burn.

Apply the patch, re-test, and magically watch the all-but-one script exit.


includes/DataObjects/GeoLocation.php
```diff
@@ -37,7 +41,7 @@ public function save()
 	{
 		if ($this->isNew) {
 // insert
-			$statement = $this->dbObject->prepare("INSERT INTO `geolocation` (address, data) VALUES (:address, :data);");
+			$statement = $this->dbObject->prepare("INSERT INTO `geolocation` (address, data) VALUES (:address, :data) ON DUPLICATE KEY UPDATE address = address;");
```
FastLizard4: What's the purpose of `ON DUPLICATE KEY UPDATE address = address` here?  Is your intention just to no-op instead of throwing an error if there's already an address in the table?

stwalkerster: Yep, exactly.

FastLizard4: Alright, fair enough

Issue #484, opened 2018-04-05
Move UI over to Bootstrap v4.


bardiharborow:
A minor nitpick really, but as I'm on the Bootstrap team it's a bit cringey to see v2.3.2 still in action. I'm happy to outlay the legwork here. If IE 8-9 support is critical, v3 might be a compromise.

stwalkerster:
Legacy support is reasonably critical, but only for the front-end, not so much for the internal interface. It should be possible to use the public side with the oldest browser you can edit Wikipedia with.

Having said that, I've looked at moving to Bootstrap v3 before, but given there's been some fairly intrusive changes to the grid system (all the classes have changed, and fluid boundaries have changed), and at the time I was sure there were some features we use removed - but that seems to not be the case with v4), I gave up on looking at it as too much work for too little benefit.

If we go ahead with this, we need to be aware of our purple button styling and sortable tables, both of which are bolt-ons to the current version of Bootstrap. I'm also tempted to ditch the bootstrap styling entirely on the front-end, and switch to the OOUI CSS instead.

Mdann52:
Hey,

I'm looking into getting some front-end experience, so I'm happy to try and put something together styling-wise if you think it will be useful!

stwalkerster:
I'm happy for either of you (or both!) to pick this up, I'd suggest this happens based on at least the [rbac branch](https://github.com/enwikipedia-acc/waca/tree/rbac) or [stwalkerster/otp](https://github.com/stwalkerster/waca/tree/otp) though so you've got a more sane baseline to use.

Note my points above about backwards compatibility with old browsers on the public interface - these restrictions don't apply on the internal interface.

Mdann52:
Noted - if you'd prefer, I'm happy to mock up a version based on BS4 and OOUI, and see which looks better and take it from there? (Personally, I think the front end would look better in the latter as well).

Just one final check - I'm going to use the list at [mediawiki](https://www.mediawiki.org/wiki/Compatibility#Browsers) unless there are any better suggestions! Would support need to extend back to basic, or just modern?

stwalkerster:
I'd personally prefer the internal side remain on bootstrap simply for familiarity, but feel free to give it a go. I'd say that the frontend must be usable for the basic set listed there - if you can use the core functionality of wikipedia you should be able to use the account request system.

FastLizard4:
[...]

PR #493, opened 2018-07-18
Migrate project to bootstrap 4

Mdann52:
Translate done and majority of project now on Bootstrap 4 - there will probably need to be a few styling tweaks, but base is there and (hopefully!) ok to be used.

Added Popper (dependency for Bootstrap dropdowns), and FontAwesome (to replace Bootstraps icons)

Also updated installing files to represent how this is installed nowadays.

Issue: #484


stwalkerster:
This looks generally good so far - I've yet to go through _all_ the changes in the diffs though.

You're right that we'll need to apply some styling fixes, which probably is going to cover quite a bit of the codebase. I expect most of this is either going to be general tweaks to the base bootstrap styling that we can do it all in our baseStyles.css file, or tweaks to the classes of specific elements.

A few things I've noticed just browsing the interface (don't be disheartened - it's a good start and a change this pervasive is always going to throw up lots of little things). This list mixes major and minor (and trivial) things, and is basically just in the order I've found them.
* Main requesting page (index.php) is very wide, possibly something odd going on with the styling here?
* Login page layout is overlapping and a little unclear
* Everything is huge - page titles I think excessively so as they're now 260px tall rather than a simple title (spotted on the request list (aka main) page and email management, but _not_ on welcome template management?)
* The default-styled button has vanished in places and looks like a regular link (spotted under welcome template management) - it looks weird with an icon.
* Edit forms appear as if they are on a mobile layout, with everything stretching right across the screen (email management)
* User management jump-to-user, and send reservation autocomplete broken
* Navbar requests button is always greyed out?
* Navbar is taller on login page than anywhere else
* Button layout on zoom page while unreserved is unclear
* Red error boxes (login, custom close) are white
* Custom close form styling missing

stwalkerster:
Something else to consider might be pulling the bootstrap/jquery/poppler/etc libraries in via composer - then we don't have a local copy we're forced to maintain.

Mdann52:
Hopefully fixed the above issues - the styling issues were almost universally due to me missing submodules (and the custom close was not where I expected it to be hence why it got missed! I've also had to add typeahead.js from Twitter into the project as typeahead support was removed from Bootstrap a few versions back.

PR #526, opened 2020-04-21
Upgrade bootstrap-sortable, fix CI, and a bit more.

mdaniels5757:
I probably should have split this up a little, but ¯\_(ツ)_/¯.

- Update the bootstrap-sortable library. There were breaking changes in between versions, but none that appear to affect us.
- SQL: we had "0000-00-00 00:00:00" as a default datetime/timestamp. This is not valid. Changed throughout to either "1000-01-01 00:00:00" (datetime) or "1970-01-01 00:00:01" (timestamp). Someone who knows more about databases than me should make sure I'm not missing anything on this.
- Travis CI: 
  - We currently are set up to ignore the failure of the only version we test. Fixed that one.
  - Let the database actually build by providing needed default environment variables.
  - Make minor changes to config per build config validator.
  - New phpunit version required minor changes to tests.
- Scrutinizer
  - Upgrade to new analysis engine.
  - Integrate codecov information.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,5 +1,12 @@
 inherit: true
 
+build:
+  nodes:
+    analysis:
+      tests:
+        override:
+          - php-scrutinizer-run
+
 tools:
     php_code_sniffer: true
     php_cpd: false
@@ -9,6 +16,7 @@ tools:
     php_pdepend: true
     php_analyzer: true
     sensiolabs_security_checker: true
+    external_code_coverage: true
 
 filter:
     excluded_paths:
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "lib/bootstrap-sortable"]
 	path = lib/bootstrap-sortable
-	url = https://github.com/enwikipedia-acc/bootstrap-sortable.git
+	url = https://github.com/drvic10k/bootstrap-sortable.git
 [submodule "lib/mediawiki-extensions-OAuth"]
 	path = lib/mediawiki-extensions-OAuth
 	url = https://github.com/wikimedia/mediawiki-extensions-OAuth.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,11 @@ language: php
 php:
   - 7.3
 
-services:
-  - mysql
+addons:
+  mariadb: '10.1'
 
 os: linux
-dist: xenial
+dist: bionic
 
 before_install:
   - pip install --user codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,21 @@ language: php
 
 php:
   - 7.3
-  
-matrix:
-  allow_failures:
-    - php: 7.3
+
+services:
+  - mysql
+
+os: linux
+dist: xenial
 
 before_install:
   - pip install --user codecov
 
-after_success:
-  - codecov
-
-sudo: false
-
 git:
   submodules: true
+
+env:
+  - MYSQL_HOST=127.0.0.1 MYSQL_USER=root MYSQL_SCHEMA=testdb
 
 before_script:
   - cd sql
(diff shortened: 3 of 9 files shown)


stwalkerster:
OK, a few initial comments from a tired me:
* I'm sure there was a reason we forked that library, and I can't recall what it was. I'll need to look it up and see if I can remember why. It would be much better if that library was exposed via composer, but alas we'll have to make do with it being a submodule. <sub>I hate submodules...</sub>
* Yay on the test fixes etc!
* Some of the unit tests in the newinternal branch use the "runkit" PHP extension for creating test mocks for more isolated unit tests. Unfortunately, last time I checked it only supported up to PHP5.6, hence why the PHP7 tests were marked as allowed failures. I don't think that affects the master branch though, so we should be good to re-enable them. A quick check reveals "runkit7" now exists, so that should also solve that issue.
* Database schema and data changes need to be done as a new migration script, not as changes to the `sql/db-structure.sql` and `sql/seed/` files. Essentially, those provide a point-in-time base, and changes are built as a series of migrations applied strictly in-order on top of that. It's the only sane way of ensuring the database is kept in sync with easy-to-test and easy-to-deply point-to-point upgrades. Creating a new db patch is fairly straight-forward - essentially, you can copy-paste the [template patch](https://github.com/enwikipedia-acc/waca/blob/master/sql/patches/patch00-example.sql), give it the next sequential number in the filename and the `patchversion` variable (line 25), then stick your SQL to migrate the database between the comments around line 55. *However*, I am hesitant to take schema changes at the moment, as the newinternal branch has a **ton** of work that I'm fighting (slowly) to get merged into master - I've sent a chaser on that one today, and I'd rather not renumber all the database patches in that.
* That being said, I haven't a clue why we've used `0000-00-00` as a date instead of `null`. MySQL [actually considers it "valid"](https://mariadb.com/kb/en/date/), and PHP [will accept garbage](https://phabricator.wikimedia.org/T133438#2232088) as input anyway. Sanity would indicate using null to indicate the absence of a value here. I'm happy(ish) with the unix epoch being used, but it involves using a "magic" value to test for, rather than just testing for the absence of a value.

If you're familiar with `git rebase`, it would be nice to split this PR up into the three main areas - the test fixes, the bootstrap-sortable upgrade, and the date changes. If not, don't worry and I'll probably do it on merge :)

stwalkerster:
I should have said - I'll follow up tomorrow on the comment about the bootstrap-sortable upgrade, and I'll also have a play with the database and see if I can figure out a reason not to use `null` in the date columns.

mdaniels5757:
Thanks for introducing me to that command. I've rebased into two commits -- a CI (Travis, Scrutinizer, phpunit) commit and a bootstrap-sortable commit. I've dropped the sql commits as not worth the effort (the only use I could think of would be to find the offending files/lines, but that can be done by searching for "0000-00-00 00:00:00"). Note this means that Travis will show a build failure due to an "invalid default value for 'time'".

.scrutinizer.yml
```diff
@@ -1,5 +1,12 @@
 inherit: true
 
+build:
+  nodes:
+    analysis:
+      tests:
+        override:
+          - php-scrutinizer-run
+
```
mdaniels5757, line 9: All this whole block does is enable Scrutinizer's new analysis engine.

stwalkerster:
I've made a few more changes to the TravisCI configuration which bring it more in-line with what we have in Prod - notably moving to MariaDB instead of MySQL, and forcing a specific set of sql_mode flags, which is allowing the build to pass now :smile: 

I still think we should move to use `null` instead of `0000-00-00 00:00:00`, but that's definitely one for another day.

stwalkerster:
OK, this is looking almost good to go now, there's just one more thing that we need to fix, which I'll do as part of the merge - the file paths changed in the bootstrap-sortable repository, so we need to update the relevant paths in our templates.

stwalkerster:
Pushed into master as 630372c

Issue #417, opened 2017-02-15
TTFB lag extremely high on users.php


dqwiki:
The TTFB takes about 20 seconds when loading the user management page. Either something is extremely wrong, or something has not been optimized.

![screenshot 36](https://cloud.githubusercontent.com/assets/676350/22957551/08a16b72-f2f7-11e6-8af2-90182a771b49.jpg)

stwalkerster:
Every time the user's onwikiname is shown, a check is done against the value from the OAuth identity token - if that token has expired, it is reloaded from Wikipedia.

This is mostly transparent - it adds around half a second to page load times in almost all cases. For user management though, we have to refresh this for every user in the same request.

I've had a look at this multiple times, and come to the conclusion that it needs a fair amount of work to refactor and change how we're handling oauth tokens such that we're happy to use an expired identity ticket. Unfortunately there's a tradeoff here - the identity ticket tells us if a user is blocked or not.

cyberpower678:
Multicurl?

Gesendet von Maximilian's iPhone. 
(Sent from Maximilian's iPhone.)

stwalkerster:
@cyberpower678 I'd rather not - API etiquette is [to do serial requests](https://www.mediawiki.org/wiki/API:Etiquette), so I'd rather stick to that where possible.

I'd like to move all of the OAuth code out of the User objects - it really doesn't belong there - and then use helper methods to either update the OAuth identity ticket or use cached data where it doesn't matter so much. This will probably be a case of always refreshing the ticket on login, plus a configurable maximum allowed time in tool configuration (probably 15 mins or something), such that the ticket is refreshed on an action by that user every 15 minutes while logged in. Looking at the current tickets issued by MediaWiki, they're only valid for 100 seconds - see this snippet from my current identity ticket on production:
```
    "exp": 1487165528,
    "iat": 1487165428,
```

Every other usage apart from the login and 15-minutely ticket refresh while logged in should use the cached data.

This is a pretty significant refactor, so I really don't want to do this in the master branch - it's not a simple fix.

cyberpower678:
This is similar to what my IABot tool does on tool labs.  It caches all OAuth data of the user and occasionally refreshes during the session.  Given that tool labs force ends the session in 5 minutes or so of no use, the OAuth refreshing of data isn’t really an issue.  Interface users are blocked if they’re blocked on the respective wiki.

stwalkerster:
[...]

Issue #265, opened 2016-06-11
The Great User / OAuth Refactor


stwalkerster:
- [x] Refactor the User class
  -- Pull out the OAuth settings + attributes we care about from identity ticket to a linked table
  -- Pull out all the OAuth code into helper classes
  -- Fix the loading of OAuth data constantly
- ~~Flip signup, do OAuth first, and then create the user object~~ Spit into #508

stwalkerster:
Also related: #417

Issue #207, opened 2015-12-25
Add create and close button


cyberpower678:
With accounts now tied to OAuth, having the options in settings to replace the big blue button with a blue and a green button would be nice.  The blue button is create account, and the green button is create and close with an identical drop down to the close created request button.  In addition, an additional checkbox, for custom closing is available, "Create account" which also create the account, and automatically mark the "account created" checkbox.

Error handling is required.  When the account process fails, it should throw a warning on the top of the request, and leave the request open and reserved, and preferably the create and close button gets greyed out.

stwalkerster:
If memory serves, this is blocked at the moment by some missing functionality on the API - not sure if that is still the case though.

cyberpower678:
Like what.  It's been possible to create accounts via the API, and it's easy to setup OAuth API calls.

cyberpower678:
Combine the two.

stwalkerster:
Yes, standard API createaccount I think is fine. The problems lie in antispoof and titleblacklist I believe, possibly SUL too. It's not just a simple job of getting the tool to do the account creation like you make it out to be. I'll take another look at this when I'm back at home from visiting family

cyberpower678:
Well with those, the account should be created manually anyway.  The tool looks for a success or fail response.

Gesendet von Maximilian's iPhone. 
(Sent from Maximilian's iPhone.)

FunPika:
The API does have ignoreantispoof and ignoretitleblacklist parameters now.

cyberpower678:
Perhaps I can work on integrating OAuth into the interface more fluently.  He claims it's slow, but I've been getting speedy responses from it.

FunPika:
Just going to note that if anyone actually does this, they should probably take care of #141 while they're at it. ;)

mdaniels5757:
I know you just closed this, but there could be a problem: antispoof sometimes doesn't catch things. We really shouldn't do this when #212 is still open (and I have experienced 212 myself recently).

stwalkerster:
#212 refers to the in-tool detection of spoofs.

If you try and create an account with an antispoof hit, whether it's detected or not, the autocreation code will vomit and chuck it back to the tool users for manual creation. It's not smart enough to handle antispoof hits or title blacklist hits. #212 is a pain, but I don't think it's actually an issue for this task.

stwalkerster:
[...]

Issue #141, opened 2014-03-01
Perform user welcomes via OAuth


stwalkerster:
Instead of the current welcomerbot and faking a signature, we should be using OAuth to authenticate as a specific user and perform the edit that way.

PR #527, opened 2020-04-26
OAuth refactoring and account auto-creation

stwalkerster:
# The Great OAuth Pull Request #

I'm sorry that this is another huge one.

## General release notes ##

* Refactor OAuth schema (#265 #127 #417) :
    - Pull OAuth token columns out of the user table
    - Create `oauthidentity` table which holds the relevant data from the JSON identity ticket

* Create a `credential` table which user credentials such as passwords.
    - Credentials are now managed by "credential providers", which allows for more than just passwords to be used as credentials.
    - Currently we hard-code everything to use passwords, but this adds flexibility in the future for other credential types.
    - Setting credentials must now be done through the relevant credential provider.

* Fix the ClearOAuthDataTask maintenance job to detach all accounts correctly, instead of hackily forcing it (#265)

* Add RefreshOAuthDataTask maintenance job to update all user's OAuth data (#265)

* Add background tasks (#207 #141)
    - Add a job queue to execute tasks in the background. This is implemented as a database table, and associated DataObject class.
    - Add maintenance job to periodically execute tasks from the jobqueue
    - Add a new page which shows the status of the job queue, and provides options to acknowledge failures and retry jobs.

* Add semi-automatic request creation (#207 #141)
    - On approval of a request, the tool will take over and create the account, send emails, and issue welcomes to the talk page.
    - Add two new "virtual" request statuses which indicate:
        a) if the request is pending jobqueue actions,
        b) if the request has been "hospitalised" due to jobqueue failures
    - Add 3 different types of background task:
        - `UserCreationTask` (creates the account via OAuth),
        - `BotCreationTask` (creates the account using a bot), and
        - `WelcomeUserTask` (leaves a welcome template on the created user's talk page).
    - Add a new user preference which determines which is the default mode in which a user creates an account (manual, oauth, bot). Role flags control which of these options are available to a user.

## Security-relevant changes ##
* `includes/Pages/UserAuth/PageChangePassword`
* `includes/Pages/UserAuth/PageForgotPassword`
* `includes/Pages/UserAuth/PageLogin`
* `includes/Pages/UserAuth/PageOAuthCallback`
* `includes/Pages/Registration/PageRegisterBase`
* `includes/DataObjects/Credential`
* `includes/Security/*`

### `WebRequest::setPartialLogin()` ###
This is a flag which holds the ID of a successfully-authenticated user who is not yet permitted to be fully-logged in. More specifically, this is set when a user is logged in as far as the tool is concerned, but needs to go through an OAuth approval before being allowed access to the tool.

Once the OAuth callback comes in, the session is upgraded to a fully-logged-in state.

### `AuthenticationManager` multi-stage login ###

AuthenticationManager is a new way of authenticating users across multiple authentication types. This should probably have gone in the upcoming OTP review, but alas it's here early.

This sits alongside the new `credential` table to manage all of a user's credentials. Every row in the `credential` table is a credential for a user, split by type, stage, and user.

`PasswordCredentialProvider` manages a credential of type `password`, handling everything that needs to be done with password storage, including validating a login request, and handling hashing of the password before storing it in the database. It also handles upgrading the password hash during the authentication phase if an upgrade to the storage medium is required.

AuthenticationManager supports multiple stages of login, and also multiple types of credential at each stage. Currently, this is completely unused - all users have a single type of credential (password) and a single stage of authentication. A good way of conceptually thinking about how this hangs together is to imagine a two-factor authentication workflow:

* Stage 1: Password
* Stage 2: TOTP token **or** U2F key **or** scratch token

At least one type of authentication **must** succeed for every stage. Thus, if multiple credentials are registered at the same stage, only one of them has to succeed for the stage to be passed. AuthenticationManager and the database support a theoretically unlimited number of credentials for a user and an unlimited number of stages for each user. However, the UI does not, and assumes a much stricter interpretation of what is allowed at each stage, which follows a much more traditional 2FA flow. Actual implementation of 2FA is not included in this pull request; that's coming next.


includes/DataObjects/User.php
```diff
@@ -779,213 +561,15 @@ public function isCommunityUser()
...
     /**
      * Gets a hash of data for the user to reset their password with.
      * @category Security-Critical
      * @return string
      */
     public function getForgottenPasswordHash()
     {
-        return md5($this->username . $this->email . $this->welcome_template . $this->id . $this->password);
+        // FIXME
+        return md5($this->username . $this->email . $this->welcome_template . $this->id);
```
stwalkerster, line 572: I am fully aware that this is too weak. I'm not planning on fixing it in this review, but I've raised #528 to fix it prior to go-live.

includes/Security/CredentialProviders/PasswordCredentialProvider.php
```diff
@@ -0,0 +1,63 @@
...
+
+    public function setCredential(User $user, $factor, $password)
+    {
+        $storedData = $this->getCredentialData($user->getId());
+
+        if($storedData === null){
+            $storedData = $this->createNewCredential($user);
+        }
+
+        $storedData->setData(password_hash($password, PASSWORD_BCRYPT, array('cost' => self::PASSWORD_COST)));
```
stwalkerster, line 57: We should consider `Argon2id` at some point now we're on PHP7.3, but that's not something for this PR.

includes/Pages/UserAuth/PagePreferences.php
```diff
@@ -0,0 +1,88 @@
...
+        $this->setHtmlTitle('Preferences');
+
+        $enforceOAuth = $this->getSiteConfiguration()->getEnforceOAuth();
+        $database = $this->getDatabase();
+        $user = User::getCurrent($database);
+
+        // Dual mode
+        if (WebRequest::wasPosted()) {
+            $this->validateCSRFToken();
+            $user->setWelcomeSig(WebRequest::postString('sig'));
```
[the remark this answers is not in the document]
stwalkerster: I'm going to close this per my escaping explanation on IRC :)

PR #531, opened 2020-05-01
Two-factor authentication

stwalkerster:
This pull request adds two-factor authentication to the tool, and as such completely upheaves the login infrastructure.

## Known issues
* A failed OAuth authorisation after an MFA login request leads the user back to the login MFA request, instead of marking the user as logged-in but pending OAuth. There's nothing inherently wrong with this, it's just ugly from a user experience point of view. I'm suggesting this is fixed after go-live, and I've raised #534 for this.

## General release notes
 Added credential providers for scratch tokens, TOTP, U2F, and Yubikey OTP.
 Rework the login process to support multiple authentication options.
* Add timeout, disabled, and priority fields to credential table
* Rename old `WebRequest::setPartialLogin` to indicate its use as an OAuth login flag. Added `WebRequest::setAuthPartialLogin` to indicate at which point of the login flow the user currently is.
 Add ability to delete credentials, but not password credentials
 Added a page for users to manage multi-factor credentials

## Security-sensitive changes
Practically everything - it's the nature of the change being an authentication process change.

Mostly:
* `includes/Pages/UserAuth/Login/*`
* `includes/Pages/UserAuth/MultiFactor/*`
* `includes/Security/*`

## Points for discussion
* Currently, only a password is required to disable a second-factor credential. Should we require a second-factor auth to disable a second-factor provider?

#### Login process
`LoginCredentialPageBase` manages most of this. There are specific overrides for the different credential types, but most of the logic is shared across all of them in this base class.

The flow is as follows:
* `LCPB::main()` runs, and as the page was a HTTP GET we show the sign in button by default (`showSignIn`), and call `setupPartial()`.
* `LCPB::setupPartial()` figures out if we have a partial login in progress, and if we do we make the relevant changes to account for this.
* `LCPB::setupAlternates()` is called by `setupPartial()` to give the user the option of switching between auth providers at a certain stage. Practically, this only matters for the `(yubikey, totp, scratch) <==> (u2f)` switch, because the former three share a form, and can all be submitted there. U2F requires it's own page.
* `LCPB::providerSpecificSetup()` gets called, which pushes through any setup needed for the chosen authentication page. Effectively:
  *  OTP auth (including scratch tokens, TOTP, and YubikeyOTP) just uses the otp template to request a OTP code.
  *  Password uses the password template. It also detects if the user is part way through a partial login, and redirects to the relevant page if necessary, instead of prompting for a password again, but this will not come up in Prod.
  *  U2F hides the sign in button (submitted via JS), and sets up the page JS to handle the authentication request correctly.
 The user is presented with the auth request, and HTTP POSTs the result back.
* `LCPB::main()` runs again, but hitting the other side of the if this time.
 We detect the login stage, and if it's not set we assume it's stage 1. If it's stage 1, we also won't have a username, so we use the POSTed value.
 We call the `getProviderCredentials()` method to retrieve the credentials which were POSTed to us. This varies per auth type.
* We pass the credentials directly to `AuthManager` with the user and the stage to authenticate against. This will return a success or failure.
  * On success, we call `processLoginSuccess` to actually log the user in properly. This also calls the relevant OAuth code from before to refresh identity tickets etc.
  * On failure, we bomb out with an auth failure message by throwing an exception. The exception is then caught, the partial login is cleared, and the user is told "no."
  * On success but when there are more stages ahead, we call `processJumpNextStage`, which figures out what credentials are needed for the next stage, and pushes the user to the relevant next stage using the nextPageMap to figure out what page a credential provider needs. Note that the presence of scratch tokens is discounted when determining the next stage, such that scratch tokens cannot be used as the sole second factor.

### `CredentialProviderBase::deleteCredential`
This has been added to not only delete credentials, but also to change the "factor" number of other credentials above this one if this one is the only one being deleted.

Imagine we've got the following (arbitrary, wouldn't happen in prod) scenario:
* Factor 1: Password
* Factor 2: TOTP
* Factor 3: YubiKey, U2F

The user could log in with either the set `{password, totp, yubikey}`, or the set `{password, totp, u2f}`. If the TOTP factor as deleted, then both of the factor 3 credentials would need to be renumbered to be factor 2, or the system would require a factor 2 but none would be present, and login would fail.

### Smarty modifiers
These are effectively functions that you can call on a value within the Smarty templates. They should only be used for formatting.

`demodhex` is a modifier which converts the modhex format to normal hex. Modhex is used by the YubiKey OTP provider.

`nlimplode` is a "natural-language" implode alternative. It eases sending an array of items to the UI, and having an "and" or "or" between the second-last and last items in the list.

### Multifactor auth
The general approach of the different credential types is explained here. Enrollment and disenrollment is managed by the PageMultiFactor class, but all the specific logic to validate and manage token storage is handled by the credential provider classes.

#### TOTP
Like the other credential types, this is almost entirely managed by the `TotpCredentialProvider` class. We use an external library to handle the complexities for us.

The standard `ICredentialProvider` methods do the following:
* `authenticate()` takes the submitted credentials, and tests them against the stored TOTP shared secret.
* `setCredential()` creates a new shared secret, and saves it encrypted to the database in a "disabled" state. This is to force the user to input the first code to validate success before we enforce the second factor, and sets a timeout on the credential to ensure the user validates it within an hour.

We also implement the following non-standard methods:

* `isPartiallyEnrolled` returns true if there's a disabled TOTP credential stored.
* `getSecret` returns the provisioning URL so we can present it to the user.
* `verifyEnable` verifies the provided TOTP token, and if correct enables the TOTP credential.

The shared secret is stored in the database in an encrypted form. The key for the encryption is stored in the application configuration file.

Enrollment is two-step. Firstly, the user is requested to authenticate, the user is presented with the generated shared secret and is asked to input a valid token. We use an external library to draw a QR code as an SVG.

If there are fewer than three scratch tokens remaining, the user is asked to regenerate the tokens on enrollment.

#### Scratch
Like the other credential types, this is almost entirely managed by the `ScratchTokenCredentialProvider` class.

The standard `ICredentialProvider` methods do the following:
* `authenticate()` takes the submitted token, and tests them against the stored list of tokens. If there is a match, remove the matching token from the list of valid tokens and save the new list back to the database.
* `setCredential()` creates a set of 5 new scratch tokens from a cryptographically secure RNG. These are then saved to the database with a low priority, so they are tested last.

We also implement the following non-standard methods:
* `getRemaining()` returns a count of remaining valid tokens.
* `getTokens()` returns the actual tokens which are currently valid. This should never be called except during the initialisation phase. I'd like to improve this to make it impossible to call outside the initialisation phase, and I have some ideas for how to accomplish this. I'd like to complete this change before go-live, so we don't screw up user's scratch tokens.

The tokens are stored in the database in an encrypted form. The key for the encryption is stored in the application configuration file.

Enrollment is a single step - tokens are regenerated following a successful authentication.

#### YubiKey OTP
Like the other credential types, this is almost entirely managed by the `YubikeyOtpCredentialProvider` class. We use an external library and external API to handle the complexities for us.

The standard `ICredentialProvider` methods do the following:
* `authenticate()` takes the submitted token, compares the first few characters of the provided credential to the stored data. This equates to the YubiKey serial number, so is a very quick determination of whether it's the same device or not. We then do an API call to the Yubico API to validate that the OTP provided is a) valid for the device, and b) generated after the last seen OTP from that device - this provides validation of the OTP and device, and prevents replay attacks.
* `setCredential()` takes the submitted token, and validates it against the Yubico API. If the token is valid, we store the serial number of the device.

We also implement the following non-standard methods:
* `getYubikeyData()` returns the serial number of the Yubikey.

The device serial number is stored in the database in plain-text. There is no secret information here which cannot be disclosed - to successfully authenticate you must be in possession of the correct hardware device.

Enrollment is a single-step - the user authenticates and presents an OTP generated by their device.

If there are fewer than three scratch tokens remaining, the user is asked to regenerate the tokens on enrollment.

#### Universal Second Factor (U2F)
Like the other credential types, this is almost entirely managed by the `U2FCredentialProvider class`. We use an external library to handle the complexities for us.

The standard `ICredentialProvider` methods do the following:
* `authenticate()` takes the submitted data and the data stored in the database, and passes both to the U2F library which handles the authentication request. If an exception is thrown, the authentication was unsuccessful.
* `setCredential()` creates a new, disabled credential with no data and a timeout. This is to allow the multi-stage enrollment to take place.

We also implement the following non-standard methods:
* `isPartiallyEnrolled` returns true if there's a disabled U2F credential stored.
* `enable()` runs the enrollment and stores the enrollment data in the database.
* `getRegistrationData()` used as part of the enrollment process
* `getAuthenticationData()` returns the data required for the U2F authentication flow to be presented to the user.

The device enrollment information is stored in the database in plain-text. There is no secret information here which cannot be disclosed - to successfully authenticate you must be in possession of the correct hardware device.

Enrollment is two-step. First the user is asked to authenticate, and then the user is requested to present their U2F token.

If there are fewer than three scratch tokens remaining, the user is asked to regenerate the tokens on enrollment.

#### Password
Like the other credential types, this is almost entirely managed by the `PasswordCredentialProvider` class.

This hasn't significantly changed, but I'm mentioning it here anyway.

The password is stored in the database in a hashed form using 10 rounds of bcrypt.

Enrollment is automatic during registration, disenrollment is not permitted. Password change is done by a separate page as before.


includes/Pages/UserAuth/Login/LoginCredentialPageBase.php
```diff
@@ -0,0 +1,335 @@
...
+            $database = $this->getDatabase();
+            try {
+                list($partialId, $partialStage) = WebRequest::getAuthPartialLogin();
+
+                if ($partialStage === null) {
+                    $partialStage = 1;
+                }
+
+                if ($partialId === null) {
+                    $username = WebRequest::postString('username');
```
dqwiki, line 61: I've looked at this function and it just grabs the user's input. There is no injection sanitation as far as I can see, even when looking at the the getByUsername function below.

includes/Pages/UserAuth/Login/LoginCredentialPageBase.php
```diff
@@ -0,0 +1,335 @@
...
+        if (!$this->enforceHttps()) {
+            return;
+        }
+
+        if (WebRequest::wasPosted()) {
+            $this->validateCSRFToken();
+
+            $database = $this->getDatabase();
+            try {
+                list($partialId, $partialStage) = WebRequest::getAuthPartialLogin();
```
dqwiki, line 54: In WebRequest.php, I don't see getAuthPartialLogin, but I see getPartialLogin - am I missing something?

includes/Pages/UserAuth/Login/LoginCredentialPageBase.php
```diff
@@ -0,0 +1,335 @@
...
+                    }
+
+                    $user = User::getByUsername($username, $database);
+                }
+                else {
+                    $user = User::getById($partialId, $database);
+                }
+
+                if ($user === false) {
+                    throw new ApplicationLogicException("Authentication failed");
```
dqwiki, line 74: This could really use a more descriptive message. I assume the only use case is when the user is not found?

includes/Pages/UserAuth/Login/LoginCredentialPageBase.php
```diff
@@ -0,0 +1,335 @@
...
+        if ($partialId !== null) {
+            // Yes, enforce this username
+            $this->partialUser = User::getById($partialId, $database);
+            $username = $this->partialUser->getUsername();
+
+            $this->setupAlternates($this->partialUser, $partialStage, $database);
+        }
+        else {
+            // No, see if we've preloaded a username
+            $preloadUsername = WebRequest::getString('tplUsername');
```
dqwiki, line 172: I haven't tracked where this comes from, but another potential injection point

includes/Pages/UserAuth/Login/LoginCredentialPageBase.php
```diff
@@ -0,0 +1,335 @@
...
+    /**
+     * @param User        $user
+     * @param int         $partialStage
+     * @param PdoDatabase $database
+     *
+     * @throws ApplicationLogicException
+     */
+    private function processJumpNextStage(User $user, $partialStage, PdoDatabase $database)
+    {
+        WebRequest::setAuthPartialLogin($user->getId(), $partialStage + 1);
```
dqwiki, line 269: Again, I don't see this function in WebRequest

includes/Pages/UserAuth/Login/LoginCredentialPageBase.php
```diff
@@ -0,0 +1,335 @@
...
+        }
+
+        $userOptions = array();
+        if (get_called_class() === PageOtpLogin::class) {
+            $userOptions = $this->setupUserOptionsForType($types, 'u2f', $userOptions);
+        }
+
+        if (get_called_class() === PageU2FLogin::class) {
+            $userOptions = $this->setupUserOptionsForType($types, 'otp', $userOptions);
+        }
```
dqwiki, line 309: Can you reassure me that this will never have both of them be false, and that the other providers don't have classes? (I couldn't find the file)

includes/Pages/UserAuth/Login/PageOtpLogin.php
```diff
@@ -0,0 +1,30 @@
...
+class PageOtpLogin extends LoginCredentialPageBase
+{
+    protected function providerSpecificSetup()
+    {
+        $this->setTemplate('login/otp.tpl');
+    }
+
+    protected function getProviderCredentials()
+    {
+        $otp = WebRequest::postString("otp");
```
dqwiki, line 23: Looks like another injection point.

includes/Pages/UserAuth/MultiFactor/PageMultiFactor.php
```diff
@@ -0,0 +1,429 @@
...
+            $this->getSiteConfiguration(), $this->getHttpHelper());
+
+        if (WebRequest::wasPosted()) {
+            $this->validateCSRFToken();
+
+            $passwordCredentialProvider = new PasswordCredentialProvider($database,
+                $this->getSiteConfiguration());
+
+            $password = WebRequest::postString('password');
+            $otp = WebRequest::postString('otp');
```
dqwiki, line 71: I won't add more possible injections beyond this - but flag again

includes/Pages/UserAuth/Login/LoginCredentialPageBase.php
stwalkerster: https://github.com/enwikipedia-acc/waca/pull/531/files#diff-3a240deb5987d4a5b631244cd0468918R516  ?

includes/Pages/UserAuth/Login/LoginCredentialPageBase.php
stwalkerster: I don't think sanitisation is needed here.

`User::getByUsername` uses parameterised queries, so there's no risk of SQL injection.

We don't write this to the database, it's a search request, so there's no risk of stored XSS.

We don't write this back to the UI, except in the case of the preloadUsename below

includes/Pages/UserAuth/Login/LoginCredentialPageBase.php
stwalkerster: This comes straight from the query string, and is routed to templates/login/shell.tpl, where it's escaped directly for output.

dqwiki:
It has been clarified that all DB queries are sanitized through PDO. This was my main concern in the comments. Had I realized this, I wouldn't have flagged it.

includes/Pages/UserAuth/Login/LoginCredentialPageBase.php
stwalkerster: There are instances where this can be false.

LCPB::setupAlternates() is called by setupPartial() to give the user the option of switching between auth providers at a certain stage. This is user-interface only, and currently this only matters if this authentication stage has at least one credential in the set (yubikey, totp, scratch) **and** also a U2F credential. `PageOtpLogin` handles all the former factors, because they're all a simple text box. U2F is a bit more complex and requires JS.

This has the effect of adding a link on the credential request page which says something like "use u2f instead" or "use a token instead".

If neither of these if blocks matches, there's no link to ask the user for an alternate credential.

If the user somehow gets to an auth page for a credential type they don't have enabled, then when they submit the form the auth request will fail because there's nothing to compare against.

PR #533, opened 2020-05-01
Fix tests for newinternal

### LINKED ISSUE #262: Prepare tests for php7
What is says on the tin.  We need to prepare our tests for an eventual move to php7

mdaniels5757:
Resolves #262. Some tests are failing, but it looks to me like they should be.


mdaniels5757:
This commit only made Travis run the tests properly, but tests look to me like they should be failing right now. (See [Travis's logs](https://travis-ci.org/github/enwikipedia-acc/waca/builds/682107459)). I'll try to debug the tests themselves tomorrow.

PR #543, opened 2020-05-20
Second Bumper Bugfix Bonanza

stwalkerster:
This set is mostly around CSP and security.

Of note, Content-Security-Policy is now set by the tool directly, and now supports nonces.

Also of note are some role changes - mostly to allow for limited roll-out of functionality.


includes/Security/RoleConfiguration.php
```diff
@@ -204,7 +202,6 @@ class RoleConfiguration
             ),
             'RequestCreation'                    => array(
                 User::CREATION_MANUAL => self::ACCESS_ALLOW,
-                User::CREATION_OAUTH  => self::ACCESS_ALLOW,
```
dqwiki, line 207: Can you quick explain this change?

stwalkerster: There's three methods by which the tool supports account creation: by link to Special:CreateAccount, automatically using OAuth, and automatically using a bot account.

This change removes the permission flag for OAuth-based creation from everyone by default, and splits it out into a new role which can be granted to a smaller group of users initially, as part of a phased rollout. This allows us to add more and more people to OAuth creation as we get more confident with it through use. Eventually I expect to undo this change and reintegrate it with the "user" role.

If it explodes on us, I don't want lots of people's creations to sort through.

stwalkerster: In this file, I've also moved U2F to be tool root only while I get to the bottom of upgrading U2F to WebAuthn, and removed 'checkuser' as a child role of 'toolRoot' so tool roots don't automatically get the UA data unless they grant themselves the CU role as well.

PR #558, opened 2020-05-30
Minor tidyup - resource cache expiry and team updates

stwalkerster:


.mailmap
```diff
@@ -4,8 +4,10 @@ Charles Melbye <charlie@yourwiki.net>
 Chris G <chris@toolserver.org>
 Cobi <cobi@jacobicarter.com>
 DeltaQuad <deltaquadwiki@gmail.com> DeltaQuad <deltaquad@live.ca>
```
dqwiki, line 6:
```suggestion
```

stwalkerster: @dqwiki - I'm not sure your change here does what you think it does.

This file maps committer identities to other identities. Currently, you have commits to this git repo under multiple identities (as most of us do...), and this tells Git to aggregate the identities together. Removing a line from this will stop them being aggregated together under the new email - essentially re-introducing your old email to the history.

stwalkerster: in any case, this is unimportant so we can deal with it later :)

PR #560, opened 2020-06-02
Search enhancements

### LINKED ISSUE #443: Make search function distinguish between confirmed requests
To reproduce, go to https://accounts.wmflabs.org/search.php?term=197&type=IP and then sort the request state to open, there's still quite a few requests such as https://accounts.wmflabs.org/acc.php?action=zoom&id=203178 which were never confirmed via email.

### LINKED ISSUE #453: Request comment search
We should also allow searching by comments - so let's add this to newinternal. At the moment, this should be based against the `rbac` branch.

It should be a simple case of adding a new method to `RequestSearchHelper` (called `byComment`), similar to the existing `byName` method, adding a new method to `PageSearch` called similar to the existing `getNameSearchResults` method, and adding a new case to the `switch` statement in `PageSearch`'s `main` method, `validateSearchParameters`, and to the form in `templates/search/searchForm.tpl`, then test it works as expected.

If we want searching on tool user comments as well as request comments, this gets slightly more in-depth, but should be possible by changing the new `RequestSearchHelper::byComment` method to have something like this:
```php
$this->modifiersClause = 'DISTINCT';
$this->joinClause = ' INNER JOIN comment c ON origin.id = c.request';
$this->whereClause .= ' AND (origin.name LIKE ? OR c.comment LIKE ?)';
$this->parameterList[] = $name;
$this->parameterList[] = $name;
```

### LINKED ISSUE #538: Search should allow refreshes
The search page should do searches by GET not POST, and it shouldn't require CSRF protection since it's not changing data.

Change the form to do a HTTP GET, remove the CSRF code from the template and `PageSearch`, and change the if statement from `WebRequest::wasPosted()` to check one of the parameters is present.

stwalkerster:
* Allows refreshes of search results page again (fixes #538)
* Adds pagination to the search results page
* Adds support for searching by request comment (fixes #453)
* Adds capability to highlight or exclude non-email confirmed requests from search results (fixes #443)


dqwiki:
I am concerned about the implementation of 443 & 453. I haven't looked through the code to see what the case is, but 443 has privacy implications if it's handed out to all users. We've had a trove of private data left in comments before and have had to do massive cleanups. I don't want this easily searchable by tool users, but admins and developers would be ok to have this limited function as they are trusted not to abuse it. ANPIP is already signed for everyone, so no additional concerns there.

443 i'm less concerned about from a privacy standpoint and just cautious from a workflow point of view. I forget if tool users have access to non-confirmed requests, but if they don't, no additional information should be showing in the search information than what they would see if looking at the page.

All of this is also assuming there are zero changes to what data is or is not included in the search interface. If that is wrong, please let me know and describe.

stwalkerster:
So the only additional data which is shown on the search interface is a flag showing whether or not a request is email-confirmed. Notably, non email-confirmed requests were previously *always* shown if they matched the search results. Requests which have not yet been email-confirmed were [previously only visible to admins](https://github.com/enwikipedia-acc/waca/blob/rel6.8.38/zoompage.php#L32) who knew to add `ecoverride` to the URL parameters. As of v7, this is [no longer possible](https://github.com/enwikipedia-acc/waca/blob/rel7.0.0/includes/Pages/PageViewRequest.php#L49). As such, it's my belief that #443 has no privacy impact when compared to the existing code. It's probably worth locking this behind a role flag, but I'm not inclined to restrict it.

Regarding #453, I'll add role flags for all the search types, granting the existing search types to the `user` role, and the comment search to the `admin` (tool admins) role.

stwalkerster:
This is currently deployed to https://accounts-dev.wmflabs.org/rc/internal.php/search if you want to have a play around as part of the review.

stwalkerster:
Replaced by #751

PR #570, opened 2020-06-15
Another bugsquish round

### LINKED ISSUE #490: Allow admins to edit requester comments
This would be useful to redact PII instead of needing DB access.

### LINKED ISSUE #567: Investigate reports of rDNS / location lookups not working
There's reports from LuK3 of this:
> When viewing the origin IP in a reserved request it does not show any basic RDNS information. It only says "(no rdns available) 
> Location unavailable".

This is something I'd noticed on a couple of requests as well, but it didn't strike me as an issue as this sort of thing happens occasionally anyway.

### LINKED ISSUE #568: View private data on related requests doesn't behave the same way post-update
Both @ElHefWiki and LuK3 have reported this. The legacy branch reportedly allowed visibility of the private data of related requests when the current request was reserved, and a closely related issue of seeing the related requests pre-reservation.

### LINKED ISSUE #569: 2FA input fields prompt for autocomplete
Alpha3031 notes:
> The new 2FA is nice, but having the field set to type="password" means my browser tries to save every single one time code. Maybe set type="text" and autocomplete="one-time-code" instead?

stwalkerster:
Fixes:
* Another outstanding bug with the refresh OAuth script, as we've got a user who predates MediaWiki logging user registration dates (#564 again), and it turns out that the reported error is actually a problem.
* Rework the request comment storage into the comment table, allowing existing comment editing to take place (Fixes #490 - prerequisite for OAuth fix because database patches...)
* Add autocomplete hints to credential fields for better browser interop (fixes #569)
* Split of visibility of related requests into a new right, granted to users (fixes #568 - regression from legacy)
* Pass another setting from config file into SiteConfiguration object; fixes location lookups (fixes #567 )


dqwiki:
I specifically would like to test #568 in sandbox for my usability report first before code review. @stwalkerster Could you verify that the sandbox is on this version?

stwalkerster:
https://accounts-dev.wmflabs.org/rc/internal.php is now running this PR, https://accounts-dev.wmflabs.org/waca/acc.php is still running the legacy tool for comparison.

stwalkerster:
Another bug squished, which should take us to zero user-reported bugs that I'm aware of in newinternal.

I've updated the RC environment to the tip of this branch for you.

stwalkerster:
RC environment updated

dqwiki:
I still can't test these changes due to the id noticeboard triggering.

PR #571, opened 2020-06-19
Change to using the vendor phpunit, as that runs all tests.

Matthewrbowker:
Note: This will fail!  It does not fix the tests or the code, only enables them to be run.  I will follow up with test-fixing pull requests in the future.


Matthewrbowker:
Actually, I just decided to fix it... 😆

PR #574, opened 2020-06-27
Improvements to bans

stwalkerster:
* Bans can now target a combination of attributes, such as IP address *and* email address
* Useragent bans now supported
* Username, email, and UA bans are regex-aware
* IP bans are now CIDR aware
* Ban targets are now restricted visibility - non-admins can't see email addresses, and non-checkusers can't see user agents
* Reason visibility added on a per-ban basis - restrict visibility of a ban's reason to admins or CUs.
* Bans now cannot be unset unless you can see the entirety of the ban
* Ban targets no longer appear in log entries. Log entries are linked to view the specific ban instead
* Bans can now auto-defer a request to a different queue, automatically drop the request, or do nothing, in addition to blocking the request from being submitted
* Adds a checkuser-level request comment visibility check.

On the back-end:
* The `ban` table has grown substantially wider; each ban how has a column for each type, as well as new columns for the new functionality
* A new `netmask` table was added, containing CIDR prefix to subnet mask conversions, for use in address-in-range queries. This table will never be changed, and is intended to be entirely read-only.
* Bump MariaDB version requirement to 10.3.22 due to use of CHECK constraints. This also allows us window functions and common table expressions. I expect versions as low as 10.2.1 will work for now, but if we're upgrading the prod DB, we should be forcing usage of that version in the dev environments too.


stwalkerster:
I've updated the database instances on wmflabs to MariaDB 10.3 in preparation for this.

stwalkerster:
I've also just cloned the RC environment to https://accounts-dev.wmflabs.org/bans/internal.php if you want to test this without running your own test environment

templates/bans/bancomplete.tpl
```diff
@@ -1 +1 @@
-﻿Banned {$ban->getTarget()|escape} for {$ban->getReason()|escape}
\ No newline at end of file
+﻿Banned set for {$ban->getReason()|escape}
```
dqwiki, line 1: Language here seems both unclear and I am unsure where this is posted.

includes/Validation/RequestValidationHelper.php
```diff
@@ -193,44 +188,69 @@ public function validateOther()
...
         // Antispoof check
-        $this->checkAntiSpoof();
+        $this->checkAntiSpoof($request);
 
         // Blacklist check
-        $this->checkTitleBlacklist();
+        $this->checkTitleBlacklist($request);
 
-        return $errorList;
+        $bans = $this->banHelper->getBans($request);
```
dqwiki, line 207: Just wishing to verify that this goes through only the active bans without dumpster diving the code. Resolve if it does.

stwalkerster: `BanHelper` gets a list of bans which apply to supplied `Request` object.

The `getBans(Request)` call is a caching wrapper around `getBansForRequestFromDatabase(Request)`, which in turn runs a database query to find any matching bans. For most attributes, such as names, this is a simple `coalesce(:name rlike name, true)` - in other words, match the requested name to a regex stored in the ban table, and if the result is `null`, return `true`. IP ranges are a little more complex, as it involves a join to the new `netmask` table to check ranges, but it's essentially doing the same thing. We also filter where `active = 1` and `duration` isn't in the past. See [L157 of `BanHelper`](https://github.com/enwikipedia-acc/waca/blob/1cbaa6f61f118374c60d248067a9a3e7afa8c0c2/includes/Helpers/BanHelper.php#L157-L174) for the full SQL query.

We then take that list of applicable bans, and loop through them, deciding what to do based on the type of ban. If it's a "do nothing ban", we do nothing. If it's a "defer" ban, we defer it to the relevant queue. If it's a drop ban, we do the relevant drop logic with relevant log entries so it is traceable that a request was created and silently dropped according to the user. Note that "block"-type bans are handled as a pre-save validation, not as a post-save validation, so we don't need to worry about them at this stage.

templates/bans/bancomplete.tpl
stwalkerster: :facepalm:

You know, I'm not actually sure where this is used either. A code search on the filename is coming up blank, so I don't think it actually *is* used. I suspect that's why I didn't notice this horrendous grammar choice when I was testing it

dqwiki (changes requested):
- Is templates/bans/main.tpl supposed to mirror almost showbans.tpl?
- Functionality of comment saving on requests is counter intuitive giving the potential need to edit comments. You must select the special save option **after** typing your comment.
- Bans that drop or defer should be limited to /20s and /48s. I know I originally said /16s, but big ranges should not be automatically banned as we are a tool of last resort for making an account. I don't see where a drop or ban would be beneficial beyond this size.

TODO: Add 2 issues:
- One requesting Ban method to be modifiable. (Priority: Low)
- A modification interface for the rest (Priority: Wishlist)

stwalkerster:
> * Is templates/bans/main.tpl supposed to mirror almost showbans.tpl?

Hrm. I guess it was split out because it *is* different, but the differences can be handled with a boolean variable, so I've merged them.

> * Functionality of comment saving on requests is counter intuitive giving the potential need to edit comments. You must select the special save option **after** typing your comment.

Yeah, I'm not entirely sure on the UI to use here. My theory was that restricted-access comments are so rarely used that it's probably not too disruptive to have a slightly harder-to-use system than I'd ideally like. For now, I'm going to leave this as an open point and have a think about it.

> * Bans that drop or defer should be limited to /20s and /48s. I know I originally said /16s, but big ranges should not be automatically banned as we are a tool of last resort for making an account. I don't see where a drop or ban would be beneficial beyond this size, *especially* since IPs for bans aren't stored in a particularly human-friendly format in the database.

Drop or *defer*? I presume you mean drop or **block**? I'm quite tempted to still permit huge ranges, but hidden behind a tool-root-only flag on the off-chance that it is needed. It'll be easier to find a tool root to apply a relevant ban than it would be to make a config change to allow larger ranges temporarily.

I'm also applying a limit for defer/nothing actions; both of these are configurable in case we want to change the limits around at some point.

> TODO: Add 2 issues:
> 
> * One requesting Ban method to be modifiable. (Priority: Low)
> * A modification interface for the rest (Priority: Wishlist)

I presume these are TODOs for you, or are you wanting me to raise the relevant issues for later consideration?

I've also found/fixed a bug with the "Ban Name" button on the view request page... :facepalm:

templates/bans/bancomplete.tpl
stwalkerster: It's not used; removed.

stwalkerster:
OK, I've redesigned the comment save thing into something which is hopefully more intuitive.

By default, this is what you see:

image

Clicking the padlock gives you this menu, from which you can choose a relevant protection type:

image

Choosing an option other than the default "locks" the padlock, and changes the button to a different colour to indicate an option has been chosen. The chosen colour matches the hue used as a highlight for the protected rows.

image
image

Does this feel like a better design to you?

stwalkerster:
`/bans/` environment updated with latest commits

stwalkerster:
> `/bans/` environment updated with latest commits

and again.

dqwiki:
#641 and #642 have been created from my to do.

> Drop or defer? I presume you mean drop or block? I'm quite tempted to still permit huge ranges, but hidden behind a tool-root-only flag on the off-chance that it is needed. It'll be easier to find a tool root to apply a relevant ban than it would be to make a config change to allow larger ranges temporarily. I'm also happy to remove this flag from the tool root group at some point in the future, but I'm not entirely comfortable with removing it from the code entirely again.
>
> I'm also applying a limit for defer/nothing actions; both of these are configurable in case we want to change the limits around at some point.

If I remember our conversation last time, I do mean drop and block. I'm fine with them being hid behind the root flag.

dqwiki (approved):
With the opening of #643

PR #623, opened 2020-10-18
Add basic database infrastructure for multi-project support

### LINKED ISSUE #590: Basic multiproject database infrastructure
* Tables
* DataObjects
* Initial request queue seed
* Initial domain seed

### `requestqueue`
| Name | Type |
|---|---|
| id | int autoinc|
| updateverion | int |
| isdefault | int not null check(in (0,1) and only one 1 in table |
| domain | int not null fk |
| ... etc - usual flags from config |

### `requestform`
| Name | Type |
|---|---|
| id | int autoinc|
| updateverion | int |
| name | string not null |
| content  | string not null |
| domain | int not null fk |
| override_queue | int fk |

### `domain`
| Name | Type |
|---|---|
| id | int autoinc |
| updateverion | int |
| shortname | string |
| longname | string |
| wiki_articlepath | string |
| wiki_api | string |
| enabled | int |
| default_template | int |
| default_language | string |
| mailing_list | string |
| notification_target | string (though actually an int stored) |

### `userdomain`
This table provides *access* to a domain, and is distinct from `domainprefs` which only holds preferences independently of domain access.
| Name | Type |
|---|---|
| id | int autoinc |
| updateversion | int |
| user | int not null fk |
| domain | int not null fk |

---
### Testing
* [x] welcome template selection/clearing still works, deletion/editing
* [x] email template editing still works, including enable/disable
* [x] request reservation/unreservation still works
* [x] enabling/disabling 2fa works
* [x] ban/unban works
* [x] oauth identity refresh works
* [x] jobqueue ack works
* [x] user forcelogout / forceid / abortpref / createmode works
* [x] signup works

stwalkerster:
* Adds four database tables for domains, request queues, and request
  forms
* Adds migration to populate domains and queue from configuration
* Changes the datatype of a lot of integers to an unsigned int
* Drop an unneeded column from emailtemplate
* Add a couple of missing foreign keys
* Add a lot of missing check constraints (now we're on MariaDB 10.3)

This is a sizable database migration due to the number of tables this is
touching.

Closes #590

### DIFF
--- a/includes/ConsoleTasks/MigrateToDomains.php
+++ b/includes/ConsoleTasks/MigrateToDomains.php
@@ -1,0 +1,62 @@
+<?php
+/******************************************************************************
+ * Wikipedia Account Creation Assistance tool                                 *
+ *                                                                            *
+ * All code in this file is released into the public domain by the ACC        *
+ * Development Team. Please see team.json for a list of contributors.         *
+ ******************************************************************************/
+
+namespace Waca\ConsoleTasks;
+
+use Waca\DataObjects\Domain;
+use Waca\DataObjects\RequestQueue;
+use Waca\Tasks\ConsoleTaskBase;
+
+class MigrateToDomains extends ConsoleTaskBase
+{
+    public function execute()
+    {
+        $requestStates = $this->getSiteConfiguration()->getRequestStates();
+        $database = $this->getDatabase();
+
+        $domain = new Domain();
+        $domain->setEnabled(true);
+        $domain->setShortName('enwiki');
+        $domain->setLongName('English Wikipedia');
+        $domain->setWikiArticlePath($this->getSiteConfiguration()->getMediawikiScriptPath());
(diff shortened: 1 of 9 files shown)


dqwiki:
I don't think this is a review that will benefit much from my looking it over. For sign off, I would like to request that a replicated version (your choice of size of the current database) have this be tested on, and try it with looking at the interface after to make sure things are good.

stwalkerster:
@dqwiki I've cloned the RC environment to https://accounts-dev.wmflabs.org/demo/internal.php, and installed this patch onto that environment. I've also found and fixed a bug with password reset which was discovered when I tried to apply the patch to the demo environment the first time, so if you want a play around with an environment with this installed, feel free to do so, though you should notice nothing different, as this patch is almost entirely just tightening up the database, moving config settings into the database, and introducing currently unused code. :)

dqwiki (approved):
Ok, i've screwed around with several functions on the test deployment. I'm going to sign off for basic functionality checked.

PR #627, opened 2020-10-31
Add viewer for logged unhandled exceptions

stwalkerster:
This is limited to tool roots only, as it's entirely possible this contains private data. Tool roots already have access to to this information in the error log directory.

This allows a much easier way of monitoring for issues and cleaning up as appropriate without requiring to log into the server.


dqwiki (approved):
nav__canErrorLog fixed

I'm not too worried about what is displayed to tool roots because they already have access. So I security audited the grants and access levels here and everything checks out.

PR #631, opened 2020-11-06
Add force-create button

stwalkerster:
This is off the back of https://phabricator.wikimedia.org/T259721

For now, this is visible to all who can see the creation buttons - we can lock it down to whoever necessary later if enwiki decides to make this CU-only.


DannyS712:
> This is off the back of https://phabricator.wikimedia.org/T259721
> 
> For now, this is visible to all who can see the creation buttons - we can lock it down to whoever necessary later if enwiki decides to make this CU-only.

Only admins (and stewards, though I'm not sure if stewards using the rights would be allowed under the global rights policy on enwiki) have `centralauth-createlocal` so showing the button to normal users of the interface is misleading since not everyone can use the special page

stwalkerster:
> showing the button to normal users of the interface is misleading since not everyone can use the special page

Yeah probably, but we don't currently have any differentiation between enwiki admins vs enwiki non-admins. Building that into the tool will be a much larger undertaking for probably minimal benefit. For that matter, we also don't have any differentiation between checkusers and stewards either - both on-wiki groups have access to the buttons for local CU/log and loginwiki CU/log.

I expect enwiki to go one of two ways - either to limit this functionality to checkusers, or to expand this functionality to accountcreators (which would make this mostly moot).

DannyS712:
> > showing the button to normal users of the interface is misleading since not everyone can use the special page
> 
> Yeah probably, but we don't currently have any differentiation between enwiki admins vs enwiki non-admins. Building that into the tool will be a much larger undertaking for probably minimal benefit. For that matter, we also don't have any differentiation between checkusers and stewards either - both on-wiki groups have access to the buttons for local CU/log and loginwiki CU/log.
> 
> I expect enwiki to go one of two ways - either to limit this functionality to checkusers, or to expand this functionality to accountcreators (which would make this mostly moot).

Perhaps add a tooltip noting that the use is restricted for now?

dqwiki (changes requested):
I can't in good conscience open this up to anyone beyond the checkuser group without a detailed splash-warning message or separate page requiring acknowledgement of the dangers of doing this.

My fear is that tool users will ask unknowing admins (or worse unknowing admins do this themselves) and create the local account of a sock over a rangeblock, particularly a sock type block without knowing the proper details.

Additionally, I don't want anyone affected by [[Wikipedia:Advice to T-Mobile IPv6 users]] to have their local account created via ACC. The level of abuse that is seen on this range and the level that sockmasters use it to create a mobile-only sock to give them clearance around a home block presents a potentially very abusive function with this.

At minimum we can modify the onwiki messages to warn admins about the creation of it (if that even exists but usually does), but I'd rather not have regular ACC users trying to rely on it, especially as non-admins don't see the day to day difficulty of block evasion most times.

Some adaptation of the current implementation will be required so i'm flagging this as Requesting changes for now.

stwalkerster:
That's fair, but this is something which absolutely must be raised on-wiki as well - as it stands, this functionality *will* be going live to all admins fairly shortly, and it's been announced in both the [admins newsletter](https://en.wikipedia.org/wiki/Wikipedia:Administrators%27_newsletter/2020/11) and [tech news](https://meta.wikimedia.org/wiki/Tech/News/2020/45) recently, there's almost bound to be cases where this functionality will start being used by unknowing admins whether we like it or not.

dqwiki:
Just for the record, I did add a local message warning administrators about it.
https://en.wikipedia.org/w/index.php?title=MediaWiki:Centralauth-createlocal-pretext&oldid=988268913

Grammar fixes welcome.

dqwiki (approved):
Changes requested made

PR #632, opened 2020-11-09
Allow usage of identity tickets for old OAuth consumer tokens

stwalkerster:
This will ease the transition from one consumer to another as existing OAuth identities will still be considered "valid" by the application.

We permit the use of expired identity tickets in a *lot* of places in the application, and this continues this tradition. If an expired ticket can be used, it will be used instead of pulling the new token. This is existing behaviour, though this change converts preferences to use the cached ticket instead of forcing a new ticket to be retrieved (which given the "refresh" button at the bottom of preferences, this is expected behaviour anyway - the previous behaviour was IMHO a bug)

Effectively, this means that logged in users can continue to use the old OAuth consumer for the duration of their login session. Other users will continue to see the cached/expired identity ticket for the legacy consumer. OAuth cache refresh jobs will not enforce a refresh of the ticket for users using the old consumer - no further use of the old consumer is permitted, so we must stick with the data we have.

That said, as soon as a user logs in (or presses "refresh" in prefs), we will direct the user down the usual OAuth approval flow for the new consumer, at which point they get an identity ticket on the new consumer and all is well again. We just need to wait for everyone to log in after the consumer change.

Failure to apply this change will result in statistics and preferences being unavailable for all users still using the legacy consumer.

Related to #141 and #207

----

This also includes a fix to the "Skip welcoming" checkbox:

This fixes a theoretically rather obscure case (but prevalant currently) where a user has welcoming enabled, but does not have the OAuth configuration required for it. This means that welcoming is always "skipped" from the user's perspective, as the "skip" checkbox is locked on.

The form POST logic relied on checking the POSTed value to determine if the welcome should be skipped due to the value of the checkbox. However, since the checkbox was disabled, it's value is never sent to the server as part of the form submission, thus the logic "saw" an unchecked checkbox and proceeded with queueing a welcome task anyway.

This is fixed by re-running the check logic on POST as well as on GET in PageCloseRequest


stwalkerster:
OK, so testing this is a bit of a PITA.

I'm setting this up in the "rc" environment, since it doesn't require database changes etc and it'll be fairly easy to switch around to different branches if necessary.

We need two OAuth consumers:
* one which matches our current configuration in Production ("legacy consumer"): https://accounts-oauth.wmflabs.org/w/index.php/Special:OAuthManageConsumers/0d484304e5982e6ebd75012409c63fff
* one which has our required configuration for account creation et al. ("new consumer"): https://accounts-oauth.wmflabs.org/w/index.php/Special:OAuthManageConsumers/e5488e35f8f00b7124225b3318791b96

The consumer key and secret key for these are in the config.local.inc.php file already - you only need to comment/uncomment the relevant one you want to use.

#### Environment setup
Firstly, let's set up the environment in the way we want to use it.
1. Clear all existing OAuth data.  (run: `php maintenance/ClearOAuthData.php` ). Verify that this has succeeded, and that the `oauthtoken` and `oauthidentity` tables are now empty.
2. Edit the configuration to enable the "legacy" consumer, disable the legacy tokens configuration entry, and enable `$enforceOAuth`
2. Log into the tool with your account (and whatever test accounts you want to log in as - I recommend having at least one other account to check and compare with). Verify that it forces you to connect your account to the OAuth wiki, and allows you to do so successfully.
3. Log out of the tool.

#### Recreate problem
Recreate the problem we're trying to solve:
1. Disable the "legacy" consumer, and enable the "new" consumer in the configuration
2. Log in to the tool with one of your accounts.
3. Verify you get prompted to approve the "new" consumer.
4. Go into user management and go to the stats page for your account.
5. Go into user management and go to the stats page for an account you've not logged into yet.

Let's go back to the start and redo the environment setup - switch back to legacy consumer, clear data, associate your account with the legacy consumer, and log out.

#### Test fix of problem
1. Disable the legacy consumer, and enable the new consumer in the configuration
2. Enable the `$oauthLegacyTokens` configuration option, ensuring that it contains the legacy consumer token in it's list.
2. Log in to the tool with one of your accounts.
3. Verify you get prompted to approve the new consumer.
4. Go into user management and go to the stats page for your account.
5. Go into user management and go to the stats page for an account you've not logged into yet.


It's worth also trying and seeing what happens when you're still logged in while the consumer changes. Check your preferences - prior to the fix this had problems. After the fix, try pressing the "Refresh OAuth identity" button in preferences from a logged-in account still on the legacy consumer.

Feel free to also try the `maintenance/RefreshOAuthData.php` script - it should emulate a login and pull the user's latest OAuth data using their stored credentials. You should see the "Cache expires" timer advance to a later time when you call this, but it won't refresh the cache for any users on a non-active consumer as that consumer should be assumed to be disabled and thus now inactive.

#### Things to look out for:
* Grants boxes on the preferences page should show the enabled grants when you have the new consumer in place
* `oauthidentity` table has an `aud` column, which refers to the "audience" for which the identity is valid - in this case that's essentially the consumer token in use at the time. It's a good tracker to see who is on which consumer.

I don't intend to use the ClearOAuthData script during the production deployment of a new consumer; the production deployment will essentially just be the steps under "Test fix of problem".

dqwiki:
For the record...

> PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /srv/dev/rc/includes/Helpers/SearchHelpers/SearchHelperBase.php on line 278
PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /srv/dev/rc/includes/Helpers/SearchHelpers/SearchHelperBase.php on line 284
PHP Warning:  array_merge(): Expected parameter 2 to be an array, int given in /srv/dev/rc/includes/Helpers/SearchHelpers/SearchHelperBase.php on line 288

dqwiki (approved):
Going to approve. The errors that were generated in the script were inconsequential to the functioning (aka everything still got nuked) so i'll leave that up to stwalkerster to look at.

All other testing went as expected without excessive errors, and the interface was clear when an error came up.

Issue #633, opened 2020-11-18
Common email domains are flagged as uncommon if in caps


ElHefWiki:
GMAIL.COM is flagged as uncommon.
gmail.com is not.

PR #634, opened 2020-11-18
Require domains to be lowercase #633

dqwiki:


stwalkerster:
This looks good, but it will only fix the issue on the zoom page. Tool admins will still see this as wrong on the main request list page - check [RequestListData L78](https://github.com/enwikipedia-acc/waca/blob/domain-capitals/includes/Fragments/RequestListData.php#L78) as well.

stwalkerster (changes requested):
.

stwalkerster (approved):

PR #640, opened 2020-11-21
Fix cleanup of non-email-confirmed requests

stwalkerster:
The old version didn't take into account comments being left on requests either by a requester or by tool users.

This meant that these requests would sit and not be deleted by the script which runs every night to clean up non-confirmed requests. In addition, because these requests were marked as "Open", these requests wouldn't have private data cleared from them either.

Now we take the approach of deleting requester comments when there are no other constraints which would prevent a request deletion, and when there are constraints which prevent deletion, we instead mark the request as closed, and "stale" to prevent an email confirmation from succeeding. The number of requests this affects should be close to zero though, because it requires a tool admin to explicitly override the email-confirmation lockout and add a comment or affect a request in such a way which adds a log entry (or a job queue entry).

We don't actually clear private data from "stale" requests immediately either; private data will only be cleared from unconfirmed requests once they are  `MAX(email-confirmation timeout, private data purge timeout)` old, which I think is acceptable.

### DIFF
--- a/includes/ConsoleTasks/OldRequestCleanupTask.php
+++ b/includes/ConsoleTasks/OldRequestCleanupTask.php
@@ -1,4 +1,5 @@
-<?php
+<?php /** @noinspection SqlConstantCondition */
+
 /******************************************************************************
  * Wikipedia Account Creation Assistance tool                                 *
  *                                                                            *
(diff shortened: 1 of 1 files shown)


includes/ConsoleTasks/OldRequestCleanupTask.php
```diff
@@ -8,24 +9,154 @@
...
+                        -- a request matching the currently-checked comment
+                        AND comment.request = r.id
+                        -- and the request is expired
+                        AND r.date < DATE_SUB(CURRENT_TIMESTAMP(), INTERVAL :expiry DAY)
+                        -- no confirmed email address
+                        AND r.emailconfirm <> 'Confirmed'
+                        -- not already marked as stale
+                        AND r.emailconfirm <> 'Stale'
+                        -- email confirmation was requested
+                        AND r.emailconfirm <> ''
```
dqwiki, line 48: Something seems fishy about this section.
The and statements would seemingly nullify finding anything.
The first comment also doesn't seem to match the code.
It could be that I'm not firing on all cylinders though.

dqwiki (changes requested):
I have one other part I need to look over when I feel more aware of what i'm looking at. I'll do that at some point in the next 24h. Marking request changes for now because of that.

includes/ConsoleTasks/OldRequestCleanupTask.php
stwalkerster: There's four possible values for this column.

* An empty string, which indicates that email confirmation is disabled and thus the request shows immediately in the UI
* `Confirmed` which indicates the requester has clicked the confirm link in their email
* `Stale` (which is newly added by this script)
* A hex string indicating the request is still pending email-confirmation

The AND statements exclude the three static values, leaving only the hex string confirmation hash.

The empty string and `Confirmed` both show in the UI, so will be dealt with and set to closed by humans. The `Stale` value is only set at the same time this script closes a request. Once the request is closed, clearing data is picked up by the other private data stripping script, rather than being completely deleted by this script.

dqwiki (approved):
Everything seems sensibly explained.

PR #650, opened 2020-12-08
Welcome Template general tidyup

### LINKED ISSUE #588: Welcome templates
In preparation for auto-welcoming (my understanding is that that's happening soon™), I reckon we should make sure all of our welcome templates are in good order. Of particular note, The Template Formerly Known As `{{Welcome}}` has been replaced and is now `{{welcome-retro}}`, so we should probably add that. There is also at least one template that's deleted (User:Mlpearc/Accwelcome, note that I only checked a couple).

I figure we should also show users a more accurate version of each template's wikicode. For example, `{{welcome}}`'s entry on <https://accounts.wmflabs.org/internal.php/welcomeTemplates> is `{{welcome}} ~~~~`, but it appears to actually be `{{subst:welcome|newuser=y}} ~~~~`. Alternatively, you could show us users the super-secret "bot code" in addition to the current code, but I don't know if that could just be confusing to others.

Cheers,
Michael (User:Mdaniels5757)

stwalkerster:
* Fix help text on editing welcome templates to better reflect what is possible
  * Deprecate `$signature` parameter; we can actually just use `~~~~` now as there is no bot managing this
  * Deprecate `$username` parameter; this was in the help text as the creator's username, but in the code as the request username. Legacy used the creator's username, so this now points to the creator.
  * Add two new unambiguous parameters to resolve any use of the above behaviour
* Centralise the templating code for the welcome templates
* Move the hard-coding of the section name into the WelcomeTemplate dataobject, in preparation for making this editable later
* Move WikiText parsing code into MediaWikiHelper to take advantage of authenticated API requests to properly inject the viewer's real signature. This removes the WikiTextHelper class entirely.
* Various HTML titles in Welcome Template Management
* Combine edit and add Smarty templates for Welcome Templates
* Show the actual code placed on the page in the main table on wide screens; show it on the "View" page for all screens.

This, combined with some data changes I've already done in Production through the UI, should get us into a much better state to enable welcoming via OAuth.

Fixes #588


stwalkerster:
I hacked in a new jobqueue task into my dev instance to test this, with an owner-only OAuth consumer. It's not a true test, as it's not using the same jobqueue task that welcoming actually would, but it tested the same fundamental code paths. See https://en.wikipedia.org/w/index.php?title=User_talk:Stwalkerster/sandbox&action=history for the edits made.

dqwiki (approved):

PR #651, opened 2020-12-09
Fix outline-dark buttons in dark mode to be more readable

stwalkerster:
This is mostly targeting the "break reservation" button on the zoom page in dark mode, but will apply for any button styled with `btn-outline-dark` while the user is on the dark theme. `btn-dark` (for example the drop button) isn't as badly affected, so I'm leaving that alone for now.


dqwiki (approved):
Cosmetics

PR #666, opened 2021-01-20
Implement support for WebAuthn

stwalkerster:
This PR removes U2F as an option for two-factor authentication, and replaces it with it's successor WebAuthn.

WebAuthn allows for multiple tokens to be registered, and for those tokens to be either "cross-platform" (ie: portable between devices such as a YubiKey), or "platform-specific" (ie: locked to a specific device, such as using Android's lockscreen or Windows Hello).

This is really quite a complex one to digest, as it drops between JS and PHP on a fairly regular basis. I'll try and write up some documentation here regarding the implementation to better explain it, but as this is (by necessity!) touching authentication code, I'd appreciate a goooooood long look at this from someone else. If you have devices available which support WebAuthn (modern Android, modern Windows 10, or a U2F token such as a YubiKey or probably some iDevice), it's probably a very good idea to test this too - I've tested this on Chrome/Linux, Firefox/Linux, Chrome/Win10, and Chrome/Android so far. Firefox/Android is [known to not support this yet](https://caniuse.com/webauthn)


stwalkerster:
Draftifying as (realisticly) this needs work to bring it back up to where the master branch is.

stwalkerster:
Given a) the complexity of this PR, b) the apparent general reluctance to review it or keep it up-to-date, and c) considerations on changes to our authentication strategy pending the WMCS ToS updates, I'm going to drop this entirely for the time being. If we want to revisit WebAuthn at a later date (and if it's appropriate/reasonable to do so), at least this PR might act as good guidance.

PR #678, opened 2021-02-28
Add Global Username List, Check If Username is IP, Don't Email If Purged

### LINKED ISSUE #619: Prevent emails being sent for purged requests
Purged requests have their underlying data replaced with 127.0.0.1 / acc@toolserver.org. If a purged request is closed with an email being sent, this will be sent to acc@toolserver.org.

We should prevent emails from being sent if the address matches the data purge address. The relevant code is likely `PageCloseRequest::sendMail(...)` ([PageCloseRequest](https://github.com/enwikipedia-acc/waca/blob/master/includes/Pages/RequestAction/PageCloseRequest.php)), and the currently configured data purge address can be pulled from the `$this->getSiteConfiguration()` ([SiteConfiguration](https://github.com/enwikipedia-acc/waca/blob/master/includes/SiteConfiguration.php)) object

### LINKED ISSUE #624: Prevent applications with IP address as requested username
Since requested usernames can't be redacted currently and IP addresses (and things that look like IP addresses) are invalid usernames, requesters shouldn't be able to ask for them. I see potentially private information hanging out there publicly fairly permanently, especially given that our user stats pages actually are publicly accessible, showing (among other things) requested usernames of rejected requests.

Alternately, come up with a way to redact requested usernames if needed - that might be a better long-term idea if feasible (for the same reasons the option exists under revdel/oversight).

### LINKED ISSUE #671: Add a Special:Globalusers link to the interface
Currently, the ACC interface only links to Special:Listusers, which only has limited usefulness for finding potential username conflicts because it only picks up accounts that are locally attached. Adding a link to Special:Globalusers would make a useful addition.

methecooldude:
Resolves #671, resolves #624, resolves #619.
First time sending a PR in to here and I'm not the neatest coder, please go easy :)

### DIFF
--- a/includes/Pages/RequestAction/PageCloseRequest.php
+++ b/includes/Pages/RequestAction/PageCloseRequest.php
@@ -195,13 +195,14 @@ class PageCloseRequest extends RequestActionBase
         if (
         ($request->getEmail() != $this->getSiteConfiguration()->getDataClearEmail()) && 
         ($request->getIp() != $this->getSiteConfiguration()->getDataClearIp())
+        )
         {
             $requestEmailHelper = new RequestEmailHelper($this->getEmailHelper());
             $requestEmailHelper->sendMail($request, $mailText, $currentUser, $ccMailingList);
             
             $request->setEmailSent(true);
             $request->save();
-            }
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


methecooldude:
Additionally resolves #624

stwalkerster (changes requested):
Thanks! 

The code here looks fine, I just have one stylistic nitpick - the existing code uses four spaces for indentation, not tabs. Would you mind changing the tabs to spaces for consistency?

methecooldude:
> Thanks!
> 
> The code here looks fine, I just have one stylistic nitpick - the existing code uses four spaces for indentation, not tabs. Would you mind changing the tabs to spaces for consistency?

You can blame my Notepad++ for that... I'll fix it up

stwalkerster (dismissed):

stwalkerster (approved):

Issue #163, opened 2014-07-19
Grammer in confirmation email link


cyberpower678:
Once your click this link, your request will be reviewed, and you will
shortly receive a separate email with more information.  Your password
is not yet available.

your->you

PR #683, opened 2021-03-02
UX Changes

### LINKED ISSUE #617: Clean up the UX surrounding ability to reopen old/purged requests
Currently, the UI buttons are always available, and throw an exception to show the error message when the user doesn't have rights to perform the relevant action.

Consider if we should disable or hide the button entirely if this action is not allowed, rather than allowing the user to click the button and be presented with the main exception handler. I'm of the opinion that maybe the button should be disabled with a tooltip saying why, but I'm also open to hiding it entirely - opinions welcome here.

Note that the throwing of the exception should still occur, but we should prevent the user from getting to that point through the usual UI.

methecooldude:
Disable reset request button after purge, unless admin. Remove 'Create Account' section from a closed request as it's pointless.
Resolves #617

### DIFF
--- a/includes/Helpers/SearchHelpers/JobQueueSearchHelper.php
+++ b/includes/Helpers/SearchHelpers/JobQueueSearchHelper.php
@@ -23,7 +23,8 @@ class JobQueueSearchHelper extends SearchHelperBase
      *
      * @return JobQueueSearchHelper
      */
-    public static function get(PdoDatabase $database) {
+    public static function get(PdoDatabase $database) 
+    {
         $helper = new JobQueueSearchHelper($database);
         return $helper;
     }
@@ -33,7 +34,8 @@ class JobQueueSearchHelper extends SearchHelperBase
      *
      * @return $this
      */
-    public function statusIn($statuses) {
+    public function statusIn($statuses) 
+    {
         $this->inClause('status', $statuses);
 
         return $this;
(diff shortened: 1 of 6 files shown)


stwalkerster (changes requested):
I've had a bit of a play around with this on the accounts-dev/rc/ environment, and I've got a few bits of feedback:
* Styling of the disabled button (I've left an inline comment on this one, but this is personal stylistic preference rather so I'm open to arguments)
* This seems to work well for purged requests (eg [#189](https://accounts-dev.wmflabs.org/rc/internal.php/viewRequest?id=189)), so :+1: 
* This doesn't seem to work for old but non-purged requests; eg [#163](https://accounts-dev.wmflabs.org/rc/internal.php/viewRequest?id=163) (yes, I ended up live-hacking data back into this req for this test).
* There's one more reset button on the main page which I'd completely forgotten about (see below). I don't think I've used that button in... forever? I'd be tempted to leave it for now, with a view to possibly removing entirely it at a later time. I don't know if anyone else uses it.
image

methecooldude:
> * Styling of the disabled button (I've left an inline comment on this one, but this is personal stylistic preference rather so I'm open to arguments)
Didn't realize that was a thing, added the tooltip instead

> * This seems to work well for purged requests (eg [#189](https://accounts-dev.wmflabs.org/rc/internal.php/viewRequest?id=189)), so 👍
> * This doesn't seem to work for old but non-purged requests; eg [#163](https://accounts-dev.wmflabs.org/rc/internal.php/viewRequest?id=163) (yes, I ended up live-hacking data back into this req for this test).
Works for me in my standard user test account...???

> * There's one more reset button on the main page which I'd completely forgotten about (see below). I don't think I've used that button in... forever? I'd be tempted to leave it for now, with a view to possibly removing entirely it at a later time. I don't know if anyone else uses it.
Yea, in theory the last 5 closed requests would not have purged data in them... that being said, I'm aware of the efforts to make ACC more global, so maybe WP projects with less ACC queues might end up in that position. I'll leave that button alone for now

stwalkerster (approved):

Issue #603, opened 2020-08-06
Implement multiproject request queue enabled flag


stwalkerster:
Handle enabled flag, show queue if requests present but not enabled

PR #704, opened 2021-08-09
Make use of request queue UI

### LINKED ISSUE #602: Migrate multiproject request queues from config
Migration from config, Request - split out status column, start using new db table, look up default queue, log changes

stwalkerster:
Switches the tool over to use the request queues stored in the database, rather than the queues stored in configuration files

Fixes #602 #603


stwalkerster:
This is deployed to https://accounts-dev.wmflabs.org/demo/internal.php

PR #726, opened 2022-03-14
Start using the settings from Domain Management

stwalkerster:
This moves some of the configuration settings from `config.inc.php` into the interface itself.

The UI for these settings already exists - it's the "Domain Management" page. This simply moves all the usages over.

At the moment, there's a lot of "FIXME: Domains!" in the code - this is intentional as this will be sorted in a later pull request. At the moment, we're only interested in pulling the settings from the default domain, ID 1. Enabling full usage of domain settings will come later.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -27,7 +27,6 @@ $toolserver_database = "";
  * File paths etc
  */
 
-$mediawikiScriptPath = "https://en.wikipedia.org/w/index.php";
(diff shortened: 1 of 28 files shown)


stwalkerster:
Deployed to https://accounts-dev.wmflabs.org/demo/internal.php for testing

dqwiki (approved):

PR #733, opened 2022-04-06
Assign users to domains

stwalkerster:
This PR does four things:
* Adds all existing users to all existing domains. In practice, this means just enwiki, since enwiki is the only domain which exists (in production)
* Defines the current Guide URL to be per-domain
* Allows a user to choose the domain they want access to when they sign up
* Sets the current domain to a domain the user is in on login.

### DIFF
--- a/sql/patches/patch40-userdomains.sql
+++ b/sql/patches/patch40-userdomains.sql
@@ -47,9 +47,14 @@ CREATE PROCEDURE SCHEMA_UPGRADE_SCRIPT() BEGIN
         modify localdocumentation varchar(255) not null;
 
     -- Add every user to every domain. In prod, this will be exactly one domain for now.
+    -- noinspection SqlWithoutWhere
+    delete from userdomain;
+
     insert into userdomain (user, domain)
     select u.id user, d.id domain
     from user u cross join domain d;
+
+    commit;
 
     -- -------------------------------------------------------------------------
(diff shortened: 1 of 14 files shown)


sql/patches/patch40-userdomains.sql
```diff
@@ -0,0 +1,66 @@
...
+
+    -- Add every user to every domain. In prod, this will be exactly one domain for now.
+    -- noinspection SqlWithoutWhere
+    delete from userdomain;
+
+    insert into userdomain (user, domain)
+    select u.id user, d.id domain
+    from user u cross join domain d;
+
+    commit;
```
FastLizard4, line 57: Nit: Triggered by the use of lowercase keywords here. You sure you missed my code reviews? :D

FastLizard4, line 57: I don't see a transaction corresponding to this commit; did you intend to start one?

FastLizard4 (approved):
One nitpick and a question, but otherwise LGTM

sql/patches/patch40-userdomains.sql
stwalkerster: No; this is mostly habit for me from Oracle which has autocommit disabled by default. Putting an explicit commit in also covers the situation in case someone's mysql client has autocommit disabled too

PR #741, opened 2022-06-11
Use Reply-To instead of From

stwalkerster:
See [T303691](https://phabricator.wikimedia.org/T303691) for details


includes/DataObjects/Domain.php
```diff
@@ -318,17 +318,17 @@ public function setDefaultLanguage(string $defaultLanguage): void
...
 
     /**
-     * @param string $emailSender
+     * @param string $emailReplyAddress
      */
-    public function setEmailSender(string $emailSender): void
+    public function setEmailReplyAddress(string $emailReplyAddress): void
     {
-        $this->emailsender = $emailSender;
+        $this->emailsender = $emailReplyAddress;
```
FastLizard4, line 331: Should the property be renamed as well? Or would that have too many knock-on effects? I'd still argue that it's maybe a good idea, even if we have to rename the database column, to avoid confusion.

FastLizard4 (commented):
Code LGTM, just a question before approving. Note that I didn't test since I didn't feel like figuring out (right now) how to get `sendmail` installed onto the `application` Docker container and how to configure it to talk to a mailsink service running within the docker-compose environment.

includes/DataObjects/Domain.php
stwalkerster: I was trying to avoid a schema change with this as I wasn't sure if I'd have to throw it in more quickly than it ended up needing to (reordering schema changes is... messy... with the current patch architecture).

Renaming the underlying field here would require a schema change do to the way PDO::FETCH_CLASS works.

I can add a schema change for this though to tidy it up properly.

FastLizard4: Ah, okay. If it's too much trouble, lmk and I'll happily approve as-is.

stwalkerster: Patch uploaded :)

FastLizard4 (dismissed):
LGTM! Also verified by testing using my brand new Docker mail sink (which I'll be submitting a PR for shortly)

FastLizard4 (approved):
:shipit:

PR #743, opened 2022-06-28
Replace Deprecated code

methecooldude:
scssPhp has deprecated the use of 'ScssPhp\\ScssPhp\\Formatter\\Compressed' and also ->comple. Replaced them with the newer versions

### DIFF
--- a/includes/ConsoleTasks/RegenerateStylesheetsTask.php
+++ b/includes/ConsoleTasks/RegenerateStylesheetsTask.php
@@ -21,7 +21,7 @@ class RegenerateStylesheetsTask extends ConsoleTaskBase
         $scss->setImportPaths('resources/scss');
 
         if (!$this->getSiteConfiguration()->getDebuggingTraceEnabled()) {
-            $scss->setFormatter(OutputStyle::COMPRESSED);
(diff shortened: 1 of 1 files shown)


includes/ConsoleTasks/RegenerateStylesheetsTask.php
```diff
@@ -32,7 +32,7 @@ public function execute()
         foreach (['bootstrap-main', 'bootstrap-alt', 'bootstrap-auto'] as $file) {
             file_put_contents(
                 self::RESOURCES_GENERATED . '/' . $file . '.css',
-                $scss->compile('/*! Do not edit this auto-generated file! */ @import "' . $file . '";'));
+                $scss->compileString('/*! Do not edit this auto-generated file! */ @import "' . $file . '";'));
```
stwalkerster, line 35: Per comment on IRC, this looks to be missing the `->getCss()` call to actually retrieve the generated CSS - at the moment it seems to be generating an empty file.

```suggestion
                $scss->compileString('/*! Do not edit this auto-generated file! */ @import "' . $file . '";')->getCss());
```

stwalkerster (changes requested):

stwalkerster (approved):

Issue #735, opened 2022-04-14
Implement Banner Alert for Flagged Comments for ACC Admins


Dane2007:
If Flagged Comments != 0, show a banner that indicates "There are X number of flagged comments for review." (clicking will bring to that review page).  This would add a bit more visibility into this instead of manually checking that screen.

stwalkerster:
I think I also want to throw in a status/queue column to that view; at the moment it's unclear whether a request is still open or not, and there have been several instances I've seen where potentially useful info has been purged out of the comment before the request was actually handled (especially in the case of deferrals to the CU queue where the comment contains an IPv6 address)

stwalkerster:
Also noting that this will need some rights checks - it's probably fine to attach this to the main right which allows you to see PageListFlaggedComments.

Questions from me:
* Do we want to show this banner everywhere, or just on the main page? (I'd prefer just the main page since we go to it often enough)
* Do we want to make it dismissable, or permanently present when there are flagged comments? If dismissable, what triggers it to be re-shown?
* Do we want it to only alert when the flagged comments are on all requests, closed requests, or just purged requests?

I can write up a rough tech spec for this if someone else wants to pick this up? I wouldn't class it as "easy", but it should be fairly straight-forward with guidance.

ElHefWiki:
Along the same lines, something I've brought up in the past is job failures sitting in the queue. Those typically sit for days before they're noticed and dealt with. Could that be rolled in with this?

Dane2007:
> Along the same lines, something I've brought up in the past is job failures sitting in the queue. Those typically sit for days before they're noticed and dealt with. Could that be rolled in with this?

I know we have a Hospital queue, perhaps we can change the styling of it to attract more attention?

Dane2007:
> Questions from me:
> * Do we want to show this banner everywhere, or just on the main page? (I'd prefer just the main page since we go to it often enough)
Just on the main page was my thought.
> * Do we want to make it dismissable, or permanently present when there are flagged comments? If dismissable, what triggers it to be re-shown?
Permanently present.
> * Do we want it to only alert when the flagged comments are on all requests, closed requests, or just purged requests?
Flagged comments on all requests.  Any request with flagged comments should be dealt with ASAP and unflagged upon resolution.

> I can write up a rough tech spec for this if someone else wants to pick this up? I wouldn't class it as "easy", but it should be fairly straight-forward with guidance.
I'm debating whether this is within my scope or ability.

[...]

PR #746, opened 2022-07-28
Add badges to highlight concern areas

Matthewrbowker:
These badges are for the flagged comments and the job queue.  For flagged comments, any comments that have a flag are counted.  For the job queue, any failed and not acknowledged error is counted.  If we have either, it also shows a badge on the main navigation item to draw attention.

#735

This is a preview of what they look like, for admins and regular users.  
Screen Shot 2022-07-28 at 4 16 25 PM
Screen Shot 2022-07-28 at 4 15 30 PM

Note that as of right now, everyone has access to the job queue.  So, the badge will show for all users.  If we want to limit that, we can.

### DIFF
--- a/includes/Fragments/NavigationMenuAccessControl.php
+++ b/includes/Fragments/NavigationMenuAccessControl.php
@@ -109,7 +109,7 @@ trait NavigationMenuAccessControl
      *
      * @return void
      */
-    public function setUpNavBarBadges($currentUser, $database) {
+    public function setUpNavBarBadges(User $currentUser, PdoDatabase $database) {
         // Set up some variables.
         // A size of 0 causes nothing to show up on the page (checked on navigation-menu.tpl) so leaving it 0 here is fine.
         $countOfFlagged = 0;
--- a/includes/Fragments/NavigationMenuAccessControl.php
+++ b/includes/Fragments/NavigationMenuAccessControl.php
@@ -8,6 +8,10 @@
 
 namespace Waca\Fragments;
 
+use Waca\DataObjects\Comment;
+use Waca\DataObjects\JobQueue;
+use Waca\DataObjects\User;
+use Waca\Helpers\SearchHelpers\JobQueueSearchHelper;
 use Waca\Pages\PageBan;
 use Waca\Pages\PageDomainManagement;
 use Waca\Pages\PageEmailManagement;
@@ -24,6 +28,7 @@ use Waca\Pages\PageViewRequest;
 use Waca\Pages\PageWelcomeTemplateManagement;
 use Waca\Pages\Statistics\StatsMain;
 use Waca\Pages\Statistics\StatsUsers;
+use Waca\PdoDatabase;
 use Waca\Security\DomainAccessManager;
 use Waca\Security\RoleConfiguration;
 use Waca\Security\SecurityManager;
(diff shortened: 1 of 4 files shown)


templates/navigation-menu.tpl
```diff
@@ -43,10 +48,24 @@
                         <a class="dropdown-item" href="{$baseurl}/internal.php/userManagement"><i class="fas fa-user"></i> User Management</a>
                     {/if}
                     {if $nav__canJobQueue}
-                        <a class="dropdown-item" href="{$baseurl}/internal.php/jobQueue"><i class="fas fa-tools"></i> Job Queue</a>
+                        <a class="dropdown-item" href="{$baseurl}/internal.php/jobQueue">
+                            <i class="fas fa-tools"></i> Job Queue
+                            {if $nav__numJobQueueFailed > 0}
+                                <span class="badge badge-danger">
+                                        {$nav__numJobQueueFailed}
+                                    </span>
```
stwalkerster, line 56: Nitpick: something odd is going on with indentation here?

includes/Fragments/NavigationMenuAccessControl.php
```diff
@@ -92,4 +97,46 @@ protected function setupNavMenuAccess($currentUser)
...
+     *
+     * This function checks to see if a user can access the pages, and if so checks the count of problem areas.
+     * If problem areas are found, a number greater than 0 will cause the badge to show up.
+     *
+     * @param User        $currentUser The current user
+     * @param PdoDatabase $database    Database instance
+     *
+     * @return void
+     */
+    public function setUpNavBarBadges($currentUser, $database) {
```
stwalkerster, line 112: Another nitpick - we're on a modern-enough version of PHP that you can do something like this to enforce the types on the function call, which we probably should be doing everywhere we can:
```suggestion
    public function setUpNavBarBadges(User $currentUser, PdoDatabase $database) {
```

Annoyingly, this has to be applied from the the called functions before it can be applied to the calling functions (or done at the same time), which is why it's not been applied above for `setupNavMenuAccess`, but I think everything here can have the type definitions applied.

templates/navigation-menu.tpl
```diff
@@ -25,7 +25,12 @@
         {/if}
         {if $nav__canBan || $nav__canEmailMgmt || $nav__canWelcomeMgmt || $nav__canSiteNoticeMgmt || $nav__canUserMgmt || $nav__canJobQueue || $nav__canFlaggedComments || $nav__canQueueMgmt || $nav__canDomainMgmt || $nav__canErrorLog}
             <li class="nav-item dropdown">
-                <a href="#" class="nav-link dropdown-toggle" aria-haspopup="true" aria-expanded="false" data-toggle="dropdown"><i class="fas fa-wrench"></i>&nbsp;Admin</a>
+                <a href="#" class="nav-link dropdown-toggle" aria-haspopup="true" aria-expanded="false" data-toggle="dropdown">
+                    <i class="fas fa-wrench"></i>&nbsp;Admin
+                    {if $nav__numAdmin > 0}
+                        <div class="badge badge-danger">{$nav__numAdmin}</div>
+                    {/if}
+                </a>
```
stwalkerster, line 33: Just thinking - how does this work on reduced-width screens? There's newly-added whitespace here (which isn't covered with nonbreaking spaces), so I'm not sure if the browser will try to wrap the text onto multiple lines.

I'll double-check this myself in a local test environment (unless you've already checked this?) so don't worry about making changes just yet (this is more of a note for myself)

stwalkerster: Never mind; other things break on the navbar before this:
Screenshot from 2022-07-29 23-26-01

(and that's a contrived worst-case example)

stwalkerster (approved):

PR #747, opened 2022-08-02
Add IPalyzer

methecooldude:
Add link to IPalyzer which looks very useful! - Example: https://ipalyzer.com/135.181.165.222 (IP is my own server at Hetzner, not linked to any request)

### DIFF
--- a/templates/view-request/ip-links.tpl
+++ b/templates/view-request/ip-links.tpl
@@ -65,6 +65,10 @@
        href="{$baseurl}/redir.php?tool=bullseye&amp;data={$ipaddress}">
         Bullseye
     </a>
+    <a id="IPalyzer-{$index}" class="btn btn-sm btn-outline-secondary visit-tracking" target="_blank"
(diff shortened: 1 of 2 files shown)


templates/view-request/ip-links.tpl
```diff
@@ -65,6 +65,10 @@
        href="{$baseurl}/redir.php?tool=bullseye&amp;data={$ipaddress}">
         Bullseye
     </a>
+    <a id="IPalyzer-{$index}" class="btn btn-sm btn-outline-secondary visit-tracking" target="_blank"
+       href="https://ipalyzer.com/{$ipaddress}">
```
stwalkerster, line 69: Tools not in WMF production should go through the redir.php script which helps strip things like referrers so we don't leak request number <=> IP address mappings to other tools.

stwalkerster (changes requested):

stwalkerster (approved):

PR #749, opened 2022-08-04
Docker improvements

FastLizard4:
- Add a `msgbroker` service running RabbitMQ to docker-compose
- Quote port numbers in docker-compose, as recommended by the spec
- Configure and enable XDebug in the Dockerfile for the application

Tested locally, using PhpStorm's auto-debug functionality (no configuration required; PhpStorm autodetects) and by creating an account request and observing a message enqueued in RabbitMQ (though I had to manually create the exchange, queue, and bindings in the RabbitMQ management interface).

Note that the `rabbitmq.conf` file is deliberately empty, but exists is a placeholder if we want to tweak the behavior of the RabbitMQ instance in `msgbroker`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,4 +20,7 @@ VOLUME /var/www/html/errorlog
 WORKDIR /var/www/html
 COPY . /var/www/html
 
+ENV XDEBUG_CONFIG="client_host=host.docker.internal client_port=9003 discover_client_host=true"
+ENV XDEBUG_MODE="develop,debug,profile"
+
 ENTRYPOINT /var/www/html/docker/entrypoint.sh
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,13 +7,14 @@ services:
     volumes:
       - .:/var/www/html/
     ports:
-      - 8080:80
+      - "8080:80"
     depends_on:
       - database
+      - msgbroker
   database:
     image: mariadb:10.3
     ports:
-      - 3306:3306
+      - "3306:3306"
     environment:
       MYSQL_ROOT_PASSWORD: waca
       MYSQL_DATABASE: waca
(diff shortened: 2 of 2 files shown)


stwalkerster:
> I had to manually create the exchange, queue, and bindings in the RabbitMQ management interface

I think the default exchange (the empty string) is a "direct" exchange with automatic bindings to queues with the name specified in the message routing key. I wonder if we can configure a startup script within the rabbit container to automatically create the objects we need for the default configuration - it might just be a queue named for the default notification target for the default domain?

I also can't remember how the default configuration works in the container; it'd be really nice if all of this works out-of-the-box. But all of this is nice-to-haves separate from this PR 😄

stwalkerster (approved):

FastLizard4:
> I think the default exchange (the empty string) is a "direct" exchange with automatic bindings to queues with the name specified in the message routing key. I wonder if we can configure a startup script within the rabbit container to automatically create the objects we need for the default configuration - it might just be a queue named for the default notification target for the default domain?

Yeah, there's a way to pass in queues via the configuration to be created at startup if they don't exist (through a `definitions.json` file that'd be trivially easy to add alongside `rabbitmq.conf`). That being said, it looked like during my testing that the default routing key is `1`; is that expected?

> I also can't remember how the default configuration works in the container; it'd be really nice if all of this works out-of-the-box. But all of this is nice-to-haves separate from this PR 😄

Yeah, the default configuration probably wouldn't work because it needs the database credentials and the container names instead of `localhost` for connecting to the DB and MQ instances (though admittedly I haven't tried it, I just made the changes locally right off the bat). My initial thinking is we could have a Docker-specific `config.local.inc.php` in the `docker` folder that gets moved into place when the application container is started if you don't already have a config file. I'll come back to all this after addressing the code review backlog!

stwalkerster:
A default routing key of 1 really doesn't matter. It's meaningless to the ACC tool, as it's the MQ infrastructure and ultimately Helpmebot itself which deals with what the different routing keys mean. 

In Helpmebot, [there is a way to map arbitrary strings to channel names](https://github.com/helpmebot/helpmebot/blob/master/src/Helpmebot.AccountCreations/Configuration/account-creations.yml#L8), but if the routing key isn't specified in the map, [it'll default to being the real target of the message](https://github.com/helpmebot/helpmebot/blob/master/src/Helpmebot.AccountCreations/Services/MqNotificationService.cs#L142-L146). Helpmebot stores the list of acceptable channels in RabbitMQ by [reading from a queue](https://github.com/helpmebot/helpmebot/blob/master/src/Helpmebot.AccountCreations/Services/MqNotificationService.cs#L63-L87) which is [only bound on the allowed routing keys](https://github.com/helpmebot/helpmebot/blob/master/src/Helpmebot.AccountCreations/Services/MqNotificationService.cs#L105-L119)

Issue #259, opened 2016-03-30
Ability for admins to manually confirm account requests


Matthewrbowker:
Sometimes, users will reply to the email list rather than clicking the confirm link.  It would be useful for tool admins to be able to manually confirm these account requests rather than directing the user to jump through more hoops.  

Tagged wishlist and needs discussion.  

Prompted by https://lists.wikimedia.org/mailman/private/accounts-enwiki-l/2016-March/019610.html

FastLizard4:
Doesn't sound like something that would be too difficult to do.  I don't think we should bother trying to list every unconfirmed request - maybe allow searching of unconfirmed requests by email and/or request ID, then allowing manual confirmation from there?

Matthewrbowker:
Yes.  Or just allow viewing of a request by URL only?  After all, we already have the ID from the subject.

FastLizard4:
Well, I'm thinking we may not necessarily, such as the instances when someone just emails the list with a subject like "confirm" and a blank body - in that case, we'd only have the email address to go on.

stwalkerster:
If they just hit reply and said confirm, we have the email below so anyone
can click on that link currently.

Otherwise, yeah, search by email (add a search parameter to the search
form) and add a mark confirmed button into /viewRequest somewhere. This
will need a confirmation option, and will need to be logged as being
performed by the relevant user rather than by the requestor, and the IRC
notification will need to be updated as well.

Related, bin the email confirm setting, and add an outbound email null
route option, disabling all outbound mail since that is the most common
issue for development and the reason that flag still exists. Would simplify
the workflow significantly too.

stwalkerster:
I'm going to suggest we take the following approach to this:

* Add a checkbox to show unconfirmed requests to the search results (will likely need a refactoring of the search page to allow multiple conditions - SearchHelper infrastructure will be useful for this)
* Change the restriction on viewing unconfirmed requests to have a button to manually mark it as confirmed, with a big nasty warning second-step reminding folk why we have email confirmation and to be doubly sure this is valid before continuing
* Log manual confirmations, and ping IRC
* Add a restricted stats page which lists all unconfirmed requests (ID, username, email) for those especially-hard-to-find instances
* Remove the ability to turn off email confirmation from a tool development perspective

Opinions @FastLizard4 @Matthewrbowker ? I think if we can do this, it's an easy task for someone to complete, albeit a larger change.

Matthewrbowker:
[...]

PR #753, opened 2022-08-05
Introduce manual confirmation

Matthewrbowker:
#259 

This is only available to administrators.  Note, this depends on #751 as that's what I based my branch on.

This commit does replace the throwing of an exception with an actual page explaining that the request is not email confirmed.  If the user is an administrator, it shows a big scary banner(tm) that warns of the consequences.  Because this provides a path to manually confirm requests, I have removed the check for `$enableEmailConfirm` out of the request view.  

User view:
Screen Shot 2022-08-05 at 12 12 01 PM

Administrator view:
Screen Shot 2022-08-05 at 12 11 53 PM

After request is confirmed (note the log):
Screen Shot 2022-08-05 at 12 12 09 PM

### DIFF
--- a/includes/Helpers/LogHelper.php
+++ b/includes/Helpers/LogHelper.php
@@ -139,6 +139,7 @@ class LogHelper
         $lookup = array(
             'Reserved'            => 'reserved',
             'Email Confirmed'     => 'email-confirmed',
+            'Manually Confirmed'     => 'manually confirmed the request',
             'Unreserved'          => 'unreserved',
             'Approved'            => 'approved',
             'Suspended'           => 'suspended',
--- a/includes/Helpers/LogHelper.php
+++ b/includes/Helpers/LogHelper.php
@@ -139,7 +139,7 @@ class LogHelper
         $lookup = array(
             'Reserved'            => 'reserved',
             'Email Confirmed'     => 'email-confirmed',
-            'Manually Confirmed'     => 'manually confirmed the request',
+            'Manually Confirmed'  => 'manually confirmed the request',
             'Unreserved'          => 'unreserved',
             'Approved'            => 'approved',
             'Suspended'           => 'suspended',
@@ -198,6 +198,7 @@ class LogHelper
             "Requests" => [
                 'Reserved'            => 'reserved',
                 'Email Confirmed'     => 'email-confirmed',
+                'Manually Confirmed'  => 'manually confirmed',
                 'Unreserved'          => 'unreserved',
                 'EditComment-c'       => 'edited a comment (by comment ID)',
                 'EditComment-r'       => 'edited a comment (by request)',
--- a/includes/Pages/PageViewRequest.php
+++ b/includes/Pages/PageViewRequest.php
@@ -59,11 +59,17 @@ class PageViewRequest extends InternalPageBase
         // Shows a page if the email is not confirmed.
         if ($request->getEmailConfirm() !== 'Confirmed') {
             // Show a banner if the user can manually confirm the request
-            $viewConfirm = $this->barrierTest("main", $currentUser, PageManuallyConfirm::class);
+            $viewConfirm = $this->barrierTest(RoleConfiguration::MAIN, $currentUser, PageManuallyConfirm::class);
+
+            // If the request is purged, there's nothing to confirm!
+            if ($request->getEmail() === $this->getSiteConfiguration()->getDataClearEmail()) {
+                $viewConfirm = false;
+            }
 
             // Render
             $this->setTemplate("view-request/not-confirmed.tpl");
             $this->assign("requestId", $request->getId());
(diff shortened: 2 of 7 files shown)


includes/Helpers/LogHelper.php
```diff
@@ -139,6 +139,7 @@ public static function getLogDescription(Log $entry)
         $lookup = array(
             'Reserved'            => 'reserved',
             'Email Confirmed'     => 'email-confirmed',
+            'Manually Confirmed'     => 'manually confirmed the request',
```
stwalkerster, line 142: Nit: alignment

includes/Pages/PageViewRequest.php
```diff
@@ -55,12 +56,18 @@ protected function main()
...
-        // Test we should be able to look at this request
-        if ($config->getEmailConfirmationEnabled()) {
-            if ($request->getEmailConfirm() !== 'Confirmed') {
-                // Not allowed to look at this yet.
-                throw new ApplicationLogicException('The email address has not yet been confirmed for this request.');
-            }
+        // Shows a page if the email is not confirmed.
+        if ($request->getEmailConfirm() !== 'Confirmed') {
+            // Show a banner if the user can manually confirm the request
+            $viewConfirm = $this->barrierTest("main", $currentUser, PageManuallyConfirm::class);
```
stwalkerster, line 62: Nit: Probably better to use the constant here?

```suggestion
            $viewConfirm = $this->barrierTest(RoleConfiguration::MAIN, $currentUser, PageManuallyConfirm::class);
```

includes/Helpers/LogHelper.php
stwalkerster, line 142: We should probably add this to `getLogActions(...)` below so this log type shows up on the filters on the log page.

templates/view-request/not-confirmed.tpl
```diff
@@ -0,0 +1,30 @@
...
+            <div class="col-lg-12">
+                <form method="post" action="{$baseurl}/internal.php/viewRequest/confirm">
+                    <div class="alert alert-danger">
+                        <input type="hidden" name="request" value="{$requestId}" />
+                        {include file="security/csrf.tpl"}
+                        As a tool administrator, you may manually confirm this request.
+                        <em>This bypasses the email confirmation step, which is not recommended unless the user has replied to the email list.</em>
+                        <br />
+                        <br />
+                        <button type="submit" class="btn btn-danger">Manually Confirm This Request</button>
```
stwalkerster, line 21: Might be better to wrap this in a `<p>`, then you shouldn't need the `<br />`s?

If you want more spacing, try adding a CSS class like [`mt-3` or `mt-5`](https://getbootstrap.com/docs/4.6/utilities/spacing/) to the button?

```suggestion
                        <p>
                            As a tool administrator, you may manually confirm this request.
                            <em>This bypasses the email confirmation step, which is not recommended unless the user has replied to the email list.</em>
                        </p>
                        <button type="submit" class="btn btn-danger">Manually Confirm This Request</button>
```

includes/Pages/RequestAction/PageManuallyConfirm.php
```diff
@@ -0,0 +1,49 @@
...
+        // Retrieve the database.
+        $database = $this->getDatabase();
+
+        // Find the request
+        // This method throws exceptions if there is an error with the request.
+        $request = $this->getRequest($database);
+
+        // Mark the request as confirmed.
+        $request->setEmailConfirm("Confirmed");
+        $request->save();
```
stwalkerster, line 38: We should pass through the updateVersion too, along with the request ID. This gives us [a form of locking](https://en.wikipedia.org/wiki/Optimistic_concurrency_control), by telling the tool "the user last saw version X of this record" (which you can pass in via `$request->setUpdateVersion(...)`). If the version of the row has changed, it means there's an edit conflict, and we should get the user to browse back and check they still want to take this action. Comparing the value passed in against the value in the database and generating the error is done automatically for you by `->save()`, but you still need to tell it the updateVersion the user last saw.

PageFlagComment does this if you want an example, or let me know if you want me to walk you through the changes needed.

stwalkerster (changes requested):
This mostly looks good - a few nitpicks, but the main things are:
* Add the new log type to the filters on PageLog
* Add optimistic locking to the manual confirmation

stwalkerster:
For extra credit: do we want to completely prevent manual confirmation of requests which have had their data purged? Given we lack an email address, we can't send decline emails, or the password for a newly created account anywhere, so it seems pointless to manually confirm a purged request.

I recognise that this is scope creep; I don't mind if you want to bundle it with this review, or open a new issue to tackle later - entirely your call.

stwalkerster (approved):
Thanks! This will get merged in once #751 is merged :)

PR #755, opened 2022-08-07
Fix SearchHelper IN() clause construction

stwalkerster:
Currently, SearchHelpers will do nothing if you pass an empty list to their SQL `IN(...)` clause generation tools. This means that the IN() is never added to the query, and it's as if you'd never specified that you wanted an `IN(...)` clause at all.

This is not expected, as I'd anticipate that specifying an `IN(...)` clause and passing an empty array would indicate you expect zero results (nothing selected to return). Indeed, several usages of this function are built in this way - for example, pull all usernames for users who reserved this set of requests, but it's an empty set to retrieve *all* usernames.

We also fix one such usage which only passes the *first* result into an `IN(...)` clause, instead of all discovered results.

This issue was spotted in the MariaDB slow query log running these queries:
```
    SELECT origin.id, origin.name FROM request origin WHERE 1 = 1
    SELECT origin.id, origin.username FROM user origin WHERE 1 = 1
```

The first query examined 200k rows on each execution, and took 0.1s to look up all requests. The second only examined 2k rows in 4ms, but was executed a large number of times. After this change, this query will be culled by the optimiser into a zero-row resultset, and shouldn't hit any tables at all.

### DIFF
--- a/includes/ConsoleTasks/ClearOAuthDataTask.php
+++ b/includes/ConsoleTasks/ClearOAuthDataTask.php
@@ -8,6 +8,7 @@
 
 namespace Waca\ConsoleTasks;
 
+use PDO;
 use Waca\Helpers\OAuthUserHelper;
 use Waca\Helpers\SearchHelpers\UserSearchHelper;
 use Waca\Tasks\ConsoleTaskBase;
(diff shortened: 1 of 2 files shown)


includes/ConsoleTasks/ClearOAuthDataTask.php
```diff
@@ -19,7 +20,7 @@ public function execute()
         $database = $this->getDatabase();
 
         $users = UserSearchHelper::get($database)->inIds(
-            $database->query('SELECT user FROM oauthtoken WHERE type = \'access\'')->fetchColumn()
+            $database->query('SELECT user FROM oauthtoken WHERE type = \'access\'')->fetchAll(PDO::FETCH_COLUMN)
```
FastLizard4, line 23: Nit: [PHP documentation](https://www.php.net/manual/en/pdostatement.fetchall.php) suggests that the `$column` parameter is required when using `PDO::FETCH_COLUMN`. But it seems to work regardless!

FastLizard4 (approved):
Optional nitpick, otherwise LGTM! :shipit:

PR #756, opened 2022-08-08
Various Docker improvements

### LINKED ISSUE #750: Docker improvements
Split from comments on #749 

The Docker infrastructure should work out-of-the-box.

For this, we need (I think):
* Docker-specific configuration overrides to default things like the database settings and the AMQP settings to the containers, but allowing overrides of other (any?) settings.
* Automated configuration of the RabbitMQ instance to work with the domain targets (IMHO, this should be _any_ routing key delivered to _one_ queue, with a sensible TTL on messages in the queue so it doesn't grow indefinitely)

FastLizard4:
Various improvements to the Docker setup for local development:

- Adds a default config.local.inc.php that will be added to the root dir if you don't already have one; should work out of the box for any Docker setup
- Adds a RabbitMQ definitions file which creates a "main" exchange and queue (also specified by default in the Docker config.local.inc.php file) that receives all notifications sent by the tool. Uses a fanout exchange so that all notifications are delivered to the same "main" queue regardless of routing key.
- Adds a mailsink service to the docker-compose environment and configures the application service to use it; this allows for easy local testing of email-based functionality.

Secondarily:
- Disable XDebug's `discover_client_host` setting as this was causing more issues than it fixed (and the hard-coded host of `host.docker.internal` should work in almost all environments)
- Update my entry in team.json

Resolves #750.

### DIFF
--- a/docker/msmtprc
+++ b/docker/msmtprc
@@ -1,0 +1,11 @@
+defaults
+auth           off
+tls            off
+tls_starttls   off
+
+account        mailsink
+host           mailsink
+port           1025
+from           accounts@wmflabs.org
+
+account default : mailsink
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,12 @@
 FROM php:7.3-apache-buster
 
+ENV DEBIAN_FRONTEND="noninteractive"
+
+COPY docker/msmtprc /etc/msmtprc
+
 # VOLUME /var/www/html
 RUN apt-get update \
-    && apt-get install git unzip -q -y --no-install-recommends \
+    && apt-get install git unzip msmtp msmtp-mta -q -y --no-install-recommends \
     && apt-get clean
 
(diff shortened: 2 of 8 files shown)


docker/msmtprc
```diff
@@ -0,0 +1,11 @@
+defaults
+auth           off
+tls            off
+tls_starttls   off
+
+account        mailsink
+host           mailsink
+port           1025
+from           accounts@wmflabs.org
```
stwalkerster, line 9: Question: what's the reason for this email address to be hardcoded here?

FastLizard4: msmtp's "sendmail" mode requires this setting, which actually sets the "envelope from" used when negotiating with the SMTP server. It's independent of and shouldn't affect the `From` header in the email. Since we're only talking to our mail sink, this can actually be any value, I just figured this would be a reasonable one.

stwalkerster (approved):

PR #762, opened 2022-08-18
switched http to https in config.inc.php , SiteConfiguration.php  and redir.php's domain mode

kumaryogesh17:
switched http to https in config.inc.php , SiteConfiguration.php  and redir.php's domain mode

### DIFF
--- a/includes/SiteConfiguration.php
+++ b/includes/SiteConfiguration.php
@@ -42,12 +42,12 @@ class SiteConfiguration
     private $oauthLegacyConsumerTokens;
     private $oauthConsumerSecret;
     private $oauthIdentityGraceTime = '24 hours';
-    private $oauthMediaWikiCanonicalServer = 'http://en.wikipedia.org';
+    private $oauthMediaWikiCanonicalServer = 'https://en.wikipedia.org';
     private $xffTrustedHostsFile = '../TrustedXFF/trusted-hosts.txt';
     private $crossOriginResourceSharingHosts = array(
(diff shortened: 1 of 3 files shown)


includes/SiteConfiguration.php
```diff
@@ -42,12 +42,12 @@ class SiteConfiguration
...
     private $oauthIdentityGraceTime = '24 hours';
-    private $oauthMediaWikiCanonicalServer = 'http://en.wikipedia.org';
+    private $oauthMediaWikiCanonicalServer = 'https://en.wikipedia.org';
     private $xffTrustedHostsFile = '../TrustedXFF/trusted-hosts.txt';
     private $crossOriginResourceSharingHosts = array(
-        "http://en.wikipedia.org",
         "https://en.wikipedia.org",
-        "http://meta.wikimedia.org",
+        "https://en.wikipedia.org",
+        "https://meta.wikimedia.org",
```
stwalkerster, line 50: No need to duplicate these HTTPS urls - one instance is sufficient.

stwalkerster (changes requested):
Apart from one minor thing (see inline comment), this looks mostly good, thanks!

includes/SiteConfiguration.php
kumaryogesh17: Thankyou . I have updated. Removed duplicate Https url from SiteConfiguration.php .

stwalkerster (approved):

PR #765, opened 2022-08-26
Hospital Check + Cleanup

methecooldude:
Add a check to see if request is in the hospital queue and prevent auto-(re)creation if so, as it doesn't make sense to offer it when it failed during that exact process.
Also, clean up the code so there are not 5 of the same divs in the code. Also, I'm sure the 2 extra 'createOauth' ids should be 'createBot'

### DIFF
--- a/templates/view-request/main.tpl
+++ b/templates/view-request/main.tpl
@@ -57,27 +57,21 @@
                                         </div>
                                     {/if}
                                     {if $canOauthCreate}
-                                        {if $requestEmailSent}
-                                            <div class="col-md-12 create-button-row {if $currentUser->getCreationMode() !== 1}d-none{/if}" id="createOauth">
+                                        <div class="col-md-12 create-button-row {if $currentUser->getCreationMode() !== 1}d-none{/if}" id="createOauth">
+                                            {if $requestEmailSent}
                                                 <div class="alert alert-warning mb-0">This request has already had an email sent to the requester. Please do a custom close or fall back to manual creation.</div>
-                                            </div>
-                                        {elseif $oauthProblem}
-                                            <div class="col-md-12 create-button-row {if $currentUser->getCreationMode() !== 1}d-none{/if}" id="createOauth">
+                                            {elseif $oauthProblem}
                                                 <div class="alert alert-warning mb-0">There's an issue with your account setup. Please check your OAuth configuration and ensure you've allowed the necessary grants.</div>
-                                            </div>
-                                        {elseif is_array($spoofs) && count($spoofs) > 0}
-                                            <div class="col-md-12 create-button-row {if $currentUser->getCreationMode() !== 1}d-none{/if}" id="createOauth">
+                                            {elseif is_array($spoofs) && count($spoofs) > 0}
                                                 <div class="alert alert-warning mb-0">This request has antispoof hits and cannot be created automatically.</div>
-                                            </div>
-                                        {elseif $requestIsBlacklisted}
-                                            <div class="col-md-12 create-button-row {if $currentUser->getCreationMode() !== 1}d-none{/if}" id="createOauth">
+                                            {elseif $requestIsBlacklisted}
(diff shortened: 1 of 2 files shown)


stwalkerster:
Just noting for the record that I _am_ looking at this; I've just hit a number of bugs while testing. I'm 95% sure that they're all existing issues though, so once I've figured out why my local environment is hosed I'll start properly looking at this.

From a code-only overview, I think what you've got is fine, but I've still to take a deeper-dive.

templates/view-request/main.tpl
```diff
@@ -57,35 +57,29 @@
...
                                                 {include file="view-request/createbuttons/auto.tpl" creationMode="oauth" }
-                                            </div>
-                                        {/if}
+                                            {/if}
+                                        </div>
                                     {/if}
                                     {if $canBotCreate}
                                         {if $requestEmailSent}
-                                            <div class="col-md-12 create-button-row {if $currentUser->getCreationMode() !== 1}d-none{/if}" id="createOauth">
+                                            <div class="col-md-12 create-button-row {if $currentUser->getCreationMode() !== 1}d-none{/if}" id="createBot">
```
stwalkerster, line 78: > Also, I'm sure the 2 extra 'createOauth' ids should be 'createBot'

You're 100% correct here, but there's something else that I've spotted here - creation mode `2` is the bot creation mode, so the if statements (all three of them - I can't catch all in this suggestion) should be checking vs 2:

```suggestion
                                            <div class="col-md-12 create-button-row {if $currentUser->getCreationMode() !== 2}d-none{/if}" id="createBot">
```

Otherwise, you get silly things like this if you have the ability to use both OAuth and Bot creations, and have your default set to OAuth:

image

Granted, that's a bit of an obscure bug that we'll never hit in production with the current configuration - bot creation isn't allowed (but may be once we get multi-domain properly working)

stwalkerster: That said, if you're working on the bot create buttons too, the same cleanup for the OAuth buttons could be applied there too to keep them similarly designed?

templates/view-request/main.tpl
```diff
@@ -57,35 +57,29 @@
...
                                                 <div class="alert alert-warning mb-0">This request has antispoof hits and cannot be created automatically.</div>
-                                            </div>
-                                        {elseif $requestIsBlacklisted}
-                                            <div class="col-md-12 create-button-row {if $currentUser->getCreationMode() !== 1}d-none{/if}" id="createOauth">
+                                            {elseif $requestIsBlacklisted}
                                                 <div class="alert alert-warning mb-0">This request has title blacklist hits and cannot be created automatically.</div>
-                                            </div>
-                                        {else}
-                                            <div class="col-md-12 create-button-row {if $currentUser->getCreationMode() !== 1}d-none{/if}" id="createOauth">
+                                            {elseif $requestStatus === 'Hospital'}
```
stwalkerster, line 69: Request statuses are defined as constants in `Waca\RequestStatus`. It would be nice if we could use the constant instead of the raw string.

There's basically two ways to do this:
1) Use the constant directly in the template. Since you can't use namespaces in templates (fully-qualified only!), that will be something like `if $requestStatus === Waca\RequestStatus::HOSPITAL`
2) Do the check in the code, and create a new boolean template variable to test in this `{if}`. I'd be tempted to do this next to where `requestIsClosed` is defined in includes/Fragments/RequestData.php, method `setupBasicData`

I have no preference to what you do here - I've used both approaches in different parts of the code.

stwalkerster (changes requested):
I'm fairly happy with this, but I would like the hard-coded request status string in the template to be replaced with the value of the constant (somehow) before merging in, as that makes future maintenance easier.

I left some comments about other adjacent fixes which are entirely optional, but since you're in the area anyway we might as well look at.

methecooldude:
I agree with the hardcoded string, I knew there was a better way of doing it, just wasn't sure how. Replaced with an assigned boolean as suggested.
Also bumped that creation mode

stwalkerster (approved):
Looks good to me, thanks!

PR #773, opened 2022-10-18
Range Requests

### LINKED ISSUE #216: Feature request: Larger "Other requests from $IP" information
Just a feature request to perhaps display a subnet/CIDR for the "other requests from IP" to evaluate multiple requests coming from the same network. Could be something small like / 28, or perhaps larger such as / 24... 

Another thought, although it would be more work, would be to consider sending a query to ARIN for the CIDR of that specific network, and then displaying all incoming requests from that range. (such as using https://github.com/addiscent/ip-addr-to-arin-cidr) For performance, I would imagine quering ARIN at request creation, but then for accuracy displaying actual "other request" data live at time of viewing/handling the request. Also to minimize performance drain would be to limit the subnet to no larger than /24, and perhaps the prior 10 requests. 

Just my 2 cents.

### LINKED ISSUE #482: Other requests from this IP also matches IP addresses for which this IP is a substring of
For example, `1.2.3.4` will also return `1.2.3.45`.

Relevant code is the LIKE clause of this query:
https://github.com/enwikipedia-acc/waca/blob/7972f26f2bdc4e55244d997bb4a7a2d5e69eaed5/includes/DataObjects/Request.php#L323-L326

methecooldude:
Resolves #216 and possibly resolves #482

Show the list of requests from the same /24 range as the requesting IP, hiding the request IP itself (as that is already handled by 'Other requests from this IP address')

### DIFF
--- a/includes/Fragments/RequestData.php
+++ b/includes/Fragments/RequestData.php
@@ -235,6 +235,16 @@ trait RequestData
 
         $this->assign('requestRelatedIpRequestsCount', count($relatedIpRequests));
         $this->assign('requestRelatedIpRequests', $relatedIpRequests);
+        
+        $relatedIpRangeRequests = RequestSearchHelper::get($database)
+            ->byIp($trustedIp, true)
+            ->withConfirmedEmail()
+            ->excludingPurgedData($configuration)
+            ->excludingRequest($request->getId())
+            ->fetch();
+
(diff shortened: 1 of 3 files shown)


dqwiki (changes requested):
Essentially going to be declining this PR. It's definitely good effort, and I can see the good intentions behind it. The problem is, this can very easily extend into more headaches with people using the tool than we would like to have.

stwalkerster:
Closing per above

PR #774, opened 2022-12-13
Add = to banned username chars

stwalkerster:
Per https://phabricator.wikimedia.org/T254045


FastLizard4 (commented):
The change itself LGTM of course, but one thing I discovered during my testing is that there seems to be some inconsistent enforcement with respect to these characters in `\Waca\Validation\RequestValidationHelper::validateName()`. If you try to request a username containing `#`. `@`, `/`, `<`, `>`, `[`, `]`, `|`, `{`, or `}` you will be prevented from submitting the request at all, while `%`, `:`, and `=` will still be allowed to be submitted, presumably only to be declined by whomever handles the request. Should these missing characters also be added to the check in `validateName()`?

dqwiki (approved):

PR #775, opened 2022-12-15
Create Docker docs

FastLizard4:
Since I've done a lot of work to support Docker-based ACC setups, making it "officially supported" by writing up a proper Docker README with all the information one might need to get started with it.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,104 @@
+This page describes running a development/testing instance of ACC within Docker, which may be easier than performing a
+"bare metal" install on a web server.
+
+**Note that, currently, this is only intended for development and testing use, and this should not be used in
+production (yet).** If you're looking for the "bare metal" installation instructions, please see
+[here](../INSTALLING.md).
+
+OAuth is also not covered as part of this document, nor is it handled by the Docker ACC setup. Refer to the
+[INSTALLING](../INSTALLING.md) file for OAuth setup instructions.
+
+# Prerequisites
+* Reasonably recent version of Docker with Docker Compose support on any Docker-supported platform
+  * Tested with Docker Desktop v4.0+ on Windows and macOS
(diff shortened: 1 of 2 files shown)


docker/README.md
```diff
@@ -0,0 +1,106 @@
...
+* (Optional) A reverse-proxy/tunnel tool with TLS support for testing features that require HTTPS, such as WebAuthn
+  * For example, [ngrok](https://ngrok.com/)
+
+Note that this document assumes that you cloned this repo into a folder called "waca". If this is not the case, some
+names of things may differ from what is given below (e.g., container and volume names will be prefixed with whatever
+you named the folder instead of "waca").
+
+# Use
+To start ACC using Docker, simply navigate in your terminal to the folder where you have cloned the waca repo and run
+the command `docker compose up --detach`. This will automatically pull the containers needed to run ACC and its
```
stwalkerster, line 23: `docker compose` instead of `docker-compose`? Is this some fancy commandline shenanigans which is only present in Docker Desktop? That subcommand certainly doesn't exist on my (Linux) machine's Docker CLI, and it looks like I'm on the latest version of Docker.

```
bash $ docker compose
docker: 'compose' is not a docker command.
See 'docker --help'

bash $ docker -v
Docker version 20.10.21, build baeda1f
```

FastLizard4: Hmm, it should be present if you've followed the recommended steps for installing Docker Compose, and running as `docker compose` is the recommendation I see everywhere. Could you try the steps [here](https://github.com/docker/compose#where-to-get-docker-compose) (linking the `docker-compose` executable to one of the `cli-plugins` locations in particular) and let me know if that fixes it for you? If so, I'll add that to the steps here.

stwalkerster (dismissed):

stwalkerster (approved):

PR #821, opened 2023-07-26
Minor tweak to client hints UI for when none are available

stwalkerster:


dqwiki (approved):
Sound code wise/only a display thing.
I think we could still streamline the hint into one-line like a UA, but again, doesn't have to be immediate.

PR #856, opened 2023-09-22
Implement global roles

stwalkerster:
Roles can now be either defined as a "local" role (like tool admin or checkuser), or a "global" role (like tool root or steward).

Global roles automatically apply across all domains. Local roles only apply in the domain in which they are granted.

There is no practical difference between how a global and a local role are treated - the only difference occurs when the role is granted. If it's defined as global in RoleConfiguration, the application will ensure it is saved as a global role instead of a local role. From a data storage point of view, a global role grant simply has the `userrole.domain` column set to null.

This means it is possible to live-hack local roles to apply as global roles or vice-versa by directly manipulating the database and it'd behave as you'd otherwise expect. Please don't actually do that though.

This change also introduces a "steward" global role. Stewards have the ability to grant +steward and +checkuser to others, and have the ability to set large IP range bans and global bans. Otherwise, their privset is the same as someone with the checkuser role.

### DIFF
--- a/includes/DataObjects/UserRole.php
+++ b/includes/DataObjects/UserRole.php
@@ -20,7 +20,7 @@ class UserRole extends DataObject
     /** @var string */
     private $role;
     /** @var int */
-    private $domain;
+    private ?int $domain;
 
     /**
      * @param int         $user
@@ -31,7 +31,7 @@ class UserRole extends DataObject
      */
     public static function getForUser(int $user, PdoDatabase $database, int $domain)
     {
-        $sql = 'SELECT * FROM userrole WHERE user = :user AND domain = :domain';
+        $sql = 'SELECT * FROM userrole WHERE user = :user AND (domain IS NULL OR domain = :domain)';
         $statement = $database->prepare($sql);
         $statement->bindValue(':user', $user);
         $statement->bindValue(':domain', $domain);
@@ -109,12 +109,12 @@ class UserRole extends DataObject
         $this->role = $role;
     }
 
-    public function getDomain(): int
+    public function getDomain(): ?int
     {
         return $this->domain;
     }
 
-    public function setDomain(int $domain): void
+    public function setDomain(?int $domain): void
     {
         $this->domain = $domain;
     }
--- a/includes/DataObjects/UserRole.php
+++ b/includes/DataObjects/UserRole.php
@@ -19,7 +19,6 @@ class UserRole extends DataObject
     private $user;
     /** @var string */
     private $role;
-    /** @var int */
     private ?int $domain;
 
     /**
--- a/includes/Helpers/LogHelper.php
+++ b/includes/Helpers/LogHelper.php
@@ -31,16 +31,15 @@ use Waca\SiteConfiguration;
 class LogHelper
 {
     /**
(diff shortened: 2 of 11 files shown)


stwalkerster:
Post-review merging blocked by the Global bans PR.

includes/DataObjects/UserRole.php
```diff
@@ -20,7 +20,7 @@ class UserRole extends DataObject
     /** @var string */
     private $role;
     /** @var int */
```
FastLizard4, line 22: Hmm, can this be removed now that there's a proper type annotation?

includes/Helpers/LogHelper.php
```diff
@@ -91,49 +85,42 @@ public static function getRequestLogsWithComments($requestId, PdoDatabase $db, S
...
-     *
-     * @param Log $entry
-     *
-     * @return string
-     */
-    public static function getLogDescription(Log $entry)
+    public static function getLogDescription(Log $entry): string
     {
-        $text = "Deferred to ";
+        $text = 'Deferred to ';
```
FastLizard4, line 90: I see that you're doing this wholesale standardization of double quotes to single quotes, but there's a few files in this PR where that doesn't appear to have been done, such as `Logger.php`. Is there a reason for this? I wonder if it would be better to split off the quotes normalization into a separate PR for ease of review/consistency?

sql/patches/patch49-globalroles.sql
```diff
@@ -0,0 +1,55 @@
...
+        SIGNAL SQLSTATE '45000' SET message_text = @message_text;
+    END IF;
+
+    -- -------------------------------------------------------------------------
+    -- Developers - put your upgrade statements here!
+    -- -------------------------------------------------------------------------
+
+    ALTER TABLE userrole
+        MODIFY COLUMN domain int(10) unsigned NULL,
+        ADD COLUMN `global` tinyint(3) unsigned GENERATED ALWAYS AS (CASE WHEN domain IS NULL THEN  1 END) VIRTUAL COMMENT 'Is this globally-defined?',
```
FastLizard4, line 39: Purely out of curiosity, why length 3?

FastLizard4 (commented):
Code LGTM, just some questions before approving

includes/DataObjects/UserRole.php
stwalkerster: Yes, done

includes/Helpers/LogHelper.php
stwalkerster: I've removed the quote changes from the PR for now

sql/patches/patch49-globalroles.sql
stwalkerster: I think this is a codegen artefact.

For MySQL/MariaDB, this really doesn't matter for numeric types as they're all the same size under the hood. The number is just used to determine the column width when displayed on the CLI.

It could be a tinyint(1) as we're only ever going to have a single digit (a single bit really) in this. The contents of the column aren't even stored to disk.

FastLizard4 (dismissed):

FastLizard4 (approved):

PR #869, opened 2023-10-01
Application startup refactoring

stwalkerster:
Primarily removing initialisation code from `config.inc.php`.

### DIFF
--- a/api.php
+++ b/api.php
@@ -16,7 +16,7 @@ use Waca\Router\ApiRequestRouter;
  * THIS IS AN ENTRY POINT
  */
 
-require_once('config.inc.php');
+require_once('includes/Startup.php');
 
 global $siteConfiguration;
 $application = new WebStart($siteConfiguration, new ApiRequestRouter());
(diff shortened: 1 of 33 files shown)


api.php
```diff
@@ -16,7 +16,7 @@
  * THIS IS AN ENTRY POINT
  */
 
-require_once('config.inc.php');
+require_once('includes/Startup.php');
```
FastLizard4, line 19: Does this change or a similar change need to be made to the `index.php` files used to prevent access to internal paths, like `/docker/index.php` and `/includes/index.php`? Trying to access `http://$baseurl/includes/`, for example, results in this:

image

The redirect works as expected in ae442df2f6c8ebe0c578e60391ce5b7ea766c0f7 from `multiproject/globalroles`. I suppose this also works, but is maybe not the optimal approach? Likewise, happy for this to be fixed in a follow-up PR, so not blocking this PR.

FastLizard4 (approved):
Code LGTM, and only noticed a minor non-blocking issue.

FastLizard4 (approved):
Had to do it manually locally, but a simple rebase with no conflict resolution, plus did a local quick liveness check, so carrying over my previous approval and will merge shortly.

PR #878, opened 2023-10-28
Add skipAutoWelcome to resolve #877

### LINKED ISSUE #877: "Skip automatic welcome on account creation" does not work
To reproduce: Manually create an account on ruwiki in the ruwiki queue. Select "Skip automatic welcome on account creation", then close as "Created!".
Observe that a welcome user job is launched. This should not happen (see below).

https://github.com/enwikipedia-acc/waca/blob/5f75b59c628b392c8320d66e7b61d5ba4ae95b27/includes/Pages/RequestAction/PageCloseRequest.php#L265-L269

Example request/job ids: 337604/15628.

\* Note: it's possible that this toggle doesn't work for automatic creation, in other queues, or after manual creation on enwiki. But I haven't tried any of those, as I want welcoming in every scenario except this one (where the job will fail because there is no corresponding enwiki user).

methecooldude:
I believe these will resolve #877 per comments from stw

The third, checkReserveProtect, doesn't need adjusting as it doesn't use showConfirmation

### DIFF
--- a/includes/Pages/RequestAction/PageCloseRequest.php
+++ b/includes/Pages/RequestAction/PageCloseRequest.php
@@ -231,6 +231,7 @@ class PageCloseRequest extends RequestActionBase
         $this->assign('emailSentOverride', WebRequest::postBoolean('emailSentOverride') ? 'true' : 'false');
         $this->assign('reserveOverride', WebRequest::postBoolean('reserveOverride') ? 'true' : 'false');
         $this->assign('createOverride', WebRequest::postBoolean('createOverride') ? 'true' : 'false');
+        $this->assign('skipAutoWelcome', webRequest::postBoolean('skipAutoWelcome') ? 'true' : 'false');
 
         $this->skipAlerts();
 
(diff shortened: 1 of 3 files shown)


includes/Pages/RequestAction/PageCloseRequest.php
```diff
@@ -231,6 +231,7 @@ protected function showConfirmation(Request $request, EmailTemplate $template, $
         $this->assign('emailSentOverride', WebRequest::postBoolean('emailSentOverride') ? 'true' : 'false');
         $this->assign('reserveOverride', WebRequest::postBoolean('reserveOverride') ? 'true' : 'false');
         $this->assign('createOverride', WebRequest::postBoolean('createOverride') ? 'true' : 'false');
+        $this->assign('skipAutoWelcome', webRequest::postBoolean('skipAutoWelcome') ? 'true' : 'false');
```
stwalkerster, line 234:
```suggestion
        $this->assign('skipAutoWelcome', WebRequest::postBoolean('skipAutoWelcome') ? 'true' : 'false');
```

stwalkerster (changes requested):

includes/Pages/RequestAction/PageCloseRequest.php
methecooldude: Oh, Derp...

stwalkerster (approved):
lgtm

PR #891, opened 2023-12-29
Fix bug where checkuser-only comments were treated as visible to all users in IRC

mdaniels5757:
(I assume that's what happened at 8:09 PM Eastern today in IRC, among other times)


stwalkerster (commented):
LGTM; will approve when we're ready to merge

FastLizard4 (commented):
LGTM to me as well

stwalkerster (approved):

Issue #900, opened 2024-01-27
Super long username issues


Matthewrbowker:
ACC is encountering a wave of spam with epically long username requests, and that's causing a few odd UI issues. 
1. Right now, requests are dropped twice by our spam filters.  This lists them twice in the the "Last 5 Closed requests" section on the main page.  Once a request is dropped by a spam filter, that should stop all other actions, as a drop is a dead-end state without user intervention.
2. Detail Page: Logs are oddly compressed on the details and times, forcing some funky scrolling.
3. Other requests from the same emails and IP overlap, and force the page to have a horizontal scroll
4. The "Requested Name" overlaps the "Log" heading on the detail page.  If anything, this is the place where we need the full username shown, but perhaps we can wrap it if it's long?  

Example request: [#340704](https://accounts.wmflabs.org/internal.php/viewRequest?id=340704)

Matthewrbowker:
Copying a quote from FL4 from #903 so it doesn't get lost...

> Testing-wise, though, I noticed a potential gap in this solution in that with the "Drop" ban action, we still send a confirmation email to the requester. This is intentional in order to give the requester as little information as possible that they have been banned, but it means that it's still possible for the requester to mark their request as email confirmed, thus bypassing the change made in this PR. Indeed, about 21% of the requests behind #900 were email confirmed and would thus not be excluded from "last closed" by this change.
> 
> It's completely fair to say that it's out-of-scope for this PR and something to be fixed in a follow-up PR, by the way; there are a couple ways the gap could be fixed. Again, I'm more making sure that you're aware of this.

PR #903, opened 2024-02-08
Hide unconfirmed requests from the last 5 closed on the main page

Matthewrbowker:
This PR hides any non-confirmed request from the last 5 closed requests if email confirmation is enabled.  This is intended as a band-aid solution to hide the spam requests from the main page (where one spam request takes two slots). Please note: this PR only touches the last 5 closed. All actions are still visible in the logs so admins and users can look for any abuse and handle it.  

The only conceivable workflow right now for a request to be dropped before being email confirmed is when a ban is in place to auto-drop a request.  If a user confirms a request and then it is dropped by a tool user/admin the request will still show up on the main page since it was handled.  

This is the beginning of work for #900.


includes/Pages/PageMain.php
```diff
@@ -68,22 +68,29 @@ protected function main()
...
-    private function setupLastFiveClosedData(PdoDatabase $database, $seeAllRequests)
+    private function setupLastFiveClosedData(PdoDatabase $database, SiteConfiguration $config, $seeAllRequests)
     {
         $this->assign('showLastFive', $seeAllRequests);
         if (!$seeAllRequests) {
             return;
         }
 
+        $queryExcludeDropped = "";
+        if($config->getEmailConfirmationEnabled()) {
```
stwalkerster, line 84: Code style nit on spacing:
```suggestion
        if ($config->getEmailConfirmationEnabled()) {
```

includes/Pages/PageMain.php
```diff
@@ -68,22 +68,29 @@ protected function main()
 
     /**
      * @param PdoDatabase $database
+     * @param SiteConfiguration $config
      * @param bool        $seeAllRequests
      *
      * @internal param User $currentUser
      */
-    private function setupLastFiveClosedData(PdoDatabase $database, $seeAllRequests)
+    private function setupLastFiveClosedData(PdoDatabase $database, SiteConfiguration $config, $seeAllRequests)
```
stwalkerster, line 76: Given this is a private function and thus completely inaccessible to anything outside the class (eg, unit tests, or other classes which may (actually don't) inherit from this class, *and* that the `$config` from the caller of the function is actually sourced from `$this->getSiteConfiguration()`, is it necessary to add this as a parameter instead of just pulling the config object directly in the function?

includes/Pages/PageMain.php
```diff
@@ -68,22 +68,30 @@ protected function main()
 
     /**
      * @param PdoDatabase $database
+     * @param SiteConfiguration $config
```
stwalkerster, line 71:
```suggestion
```

stwalkerster (commented):
LGTM with one last tweak 😀

stwalkerster (commented):
LGTM; will approve when we're ready to merge

FastLizard4 (commented):
One nitpick from me, otherwise code LGTM.

Testing-wise, though, I noticed a potential gap in this solution in that with the "Drop" ban action, we still send a confirmation email to the requester. This is intentional in order to give the requester as little information as possible that they have been banned, but it means that it's still possible for the requester to mark their request as email confirmed, thus bypassing the change made in this PR. Indeed, about 21% of the requests behind #900 were email confirmed and would thus not be excluded from "last closed" by this change.

It's completely fair to say that it's out-of-scope for this PR and something to be fixed in a follow-up PR, by the way; there are a couple ways the gap could be fixed. Again, I'm more making sure that you're aware of this.

stwalkerster (approved):

PR #924, opened 2024-04-06
Fix broken statistics

mdaniels5757:
Statistics for "Tool administrators" and "Tool users" are broken right now on prod. I think this fixes them. (But I couldn't get the docker to work in order to test this, so please make sure this doesn't break anything!)

### DIFF
--- a/includes/Pages/Statistics/StatsMain.php
+++ b/includes/Pages/Statistics/StatsMain.php
@@ -75,7 +75,7 @@ SQL
         $unconfirmedStatement->closeCursor();
         $this->assign('statsUnconfirmed', $unconfirmed);
 
-        $userRoleStatement = $database->prepare('SELECT COUNT(*) FROM user INNER JOIN userrole ON user.id = userrole.user WHERE role = :role;');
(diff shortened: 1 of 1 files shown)


includes/Pages/Statistics/StatsMain.php
```diff
@@ -75,20 +75,22 @@ private function generateSmallStatsTable()
         $unconfirmedStatement->closeCursor();
         $this->assign('statsUnconfirmed', $unconfirmed);
 
-        $userStatusStatement = $database->prepare('SELECT COUNT(*) FROM user WHERE status = :status;');
+        $userRoleStatement = $database->prepare('SELECT COUNT(*) FROM user INNER JOIN userrole ON user.id = userrole.user WHERE role = :role;');
```
stwalkerster, line 78: This will include all users with those roles, whether they're active or suspended. Is this what we _want_ to show? If we only want to show active users, we'll need a where condition on user.status == Active too

stwalkerster (commented):
LGTM and seems to work fine. One point raised inline though.

includes/Pages/Statistics/StatsMain.php
mdaniels5757: I think we want to only show active users. Pushed a new commit, please take a look.

stwalkerster (approved):

PR #940, opened 2024-05-26
Relicence project to the Unlicense

stwalkerster:
This project previously used a public domain declaration, which is known to cause [difficulties][osi-pd] legally. By moving to the Unlicense, we stick with the public domain intentions as closely as possible, while still using an OSI-approved licence.

[osi-pd]: https://opensource.org/faq#public-domain


FastLizard4 (approved):
Hands-down the single most exciting PR I've ever reviewed for all time.

PR #964, opened 2024-08-12
Amqp deprecation fix

### LINKED ISSUE #957: Fix deprecation issue in php-amqplib
https://github.com/php-amqplib/php-amqplib/pull/1178/files

That change probably affects us, and we'll need to update our code to handle it. For now, it should just be a deprecation warning, but that will change at some point.

The relevant code is all in the IRC notifications code - that's the only place we use AMQP.

mdaniels5757:
Resolves #957.

Notes:
* I tested this with the Docker config's default RabbitMQ settings (which do not include SSL/TLS), and RabbitMQ received the messages correctly. I could not test with SSL/TLS because I'm running the tool on my local machine.
* I got multiple exceptions/errors when setting up, logging in, creating a request, and reserving that request:
    * `WARN[0000] /home/m/waca/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion`
    * `( ! ) Warning: Undefined array key "hidden" in /var/www/html/templates_c/d5b8cd18de789ef98bed0c224d77de2b5c511f8b_0.file.request-log.tpl.php on line 48` in the comments section after reserving the request.
    * `( ! ) Deprecated: Using unregistered function "count" in a template is deprecated and will be removed in a future release. Use Smarty::registerPlugin to explicitly register a custom modifier. in /var/www/html/vendor/smarty/smarty/libs/sysplugins/smarty_internal_templatecompilerbase.php on line 663` at multiple points.

Are any of those known, or should I create (and maybe work on) separate issues for them?

### DIFF
--- a/includes/Helpers/IrcNotificationHelper.php
+++ b/includes/Helpers/IrcNotificationHelper.php
@@ -115,14 +115,15 @@ class IrcNotificationHelper
             else {
                 $amqpConnectionConfig->setIsSecure(false);
             }
-            $connection = AMQPConnectionFactory::create($amqpConnectionConfig);
 
+            $connection = AMQPConnectionFactory::create($amqpConnectionConfig);
             $channel = $connection->channel();
 
             $msg = new AMQPMessage(substr($msg, 0, 512));
             $msg->set('user_id', $amqpSiteConfig['user']);
             $msg->set('app_id', $this->siteConfiguration->getUserAgent());
             $msg->set('content_type', 'text/plain');
+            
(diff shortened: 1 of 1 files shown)


methecooldude:
I have raised PR #965 in regards to the docker warning and the Smarty deprecation

stwalkerster:
> I could not test with SSL/TLS because I'm running the tool on my local machine.

I'm 90% sure that we're not running RabbitMQ with TLS in production either (localhost-only)

stwalkerster (commented):
LGTM in principle, I'll try and test it locally in the coming days before merging 😄

stwalkerster (approved):

PR #965, opened 2024-08-12
Fixes

methecooldude:
Per Michael's PR #964. Remove version from docker-compose which is obsolete. Add count as a Smarty plugin so it continues working if Smarty is updated
EDIT: Just realised the Smarty issue is part of #794


includes/Fragments/TemplateOutput.php
```diff
@@ -44,6 +44,8 @@ final protected function setUpSmarty()
     {
         $this->smarty = new Smarty();
         $this->smarty->addPluginsDir($this->getSiteConfiguration()->getFilePath() . '/smarty-plugins');
+        $this->registerPlugin('modifier', 'count', 'count');
+        $this->registerPlugin('modifier', 'implode', 'implode');
```
stwalkerster, line 48: Is this supposed to be `$this->smarty->registerPlugin(...)`?

```suggestion
        $this->smarty->registerPlugin('modifier', 'implode', 'implode');
```

Looks like `count` [is supported](https://smarty-php.github.io/smarty/stable/designers/language-modifiers/language-modifier-count/) in Smarty 5.x ([and 4.5.4](https://github.com/smarty-php/smarty/issues/813#issuecomment-2270648968)), so we probably don't need to worry about that one

stwalkerster (changes requested):
One more thing (paperwork mostly!) - did you happen to check if this solves #794, or is there still more to deal with before we upgrade to 5.x? (I'm not asking you to do this, I'm just curious if we can close that issue after this PR is merged)

includes/Fragments/TemplateOutput.php
methecooldude: > Is this supposed to be `$this->smarty->registerPlugin(...)`?
Yes, yes it is, I'll re-commit that in a moment

> Looks like `count` [is supported](https://smarty-php.github.io/smarty/stable/designers/language-modifiers/language-modifier-count/) in Smarty 5.x ([and 4.5.4](https://github.com/smarty-php/smarty/issues/813#issuecomment-2270648968)), so we probably don't need to worry about that one

Not according the to deprecation notice that Michael posted in the other PR, and doing it now will save hassle later?

methecooldude:
> One more thing (paperwork mostly!) - did you happen to check if this solves #794, or is there still more to deal with before we upgrade to 5.x? (I'm not asking you to do this, I'm just curious if we can close that issue after this PR is merged)

I didn't check the rest of it, it's just you mentioned implode on the issue, and I realised that the same fix for count can be done for impode

includes/Fragments/TemplateOutput.php
stwalkerster: Yeah, we're currently on 4.5.3, and the count deprecation notice was seemingly removed in 4.5.4 as it's supported directly on 5.x

stwalkerster (approved):

includes/Fragments/TemplateOutput.php
methecooldude: > Yeah, we're currently on 4.5.3, and the count deprecation notice was seemingly removed in 4.5.4 as it's supported directly on 5.x

*sigh*

PR #969, opened 2024-08-18
Upgrade to PHP 8.2

mdaniels5757:
It looks like some dependencies are removing support for PHP 7.4 (which has been EOL for almost two years). Although none of those dependencies' new releases seem to contain security fixes, being on a supported PHP version would likely make applying future security fixes easier.

Notes:
* GitHub Actions will now also run tests for PHP 8.3, but allow them to fail if PHP 8.2 is passing. Since PHP 8.2 is what is in Debian Bookworm (which we use for our database), I assume that's what we would be using.
* This applies the upgrades in #921, #948, and #949. I tested 2FA setup (with QR code) and usage and it worked well.
* The changes from `Smarty` to `Smarty\Smarty` and similar is due to the upgrade to Smarty v5 (https://smarty-php.github.io/smarty/stable/upgrading/).
* PHPUnit 10 (#966) and 11 (#963) are compatible with PHP 8.2, but will require more changes separately.
* I made too many commits. Please "squash and merge" if you merge this. :)


stwalkerster:
Just as a heads-up, we're still running the production application on bullseye, which is still PHP 7.4. This can't be merged until we've got the new infrastructure in-place. We will indeed be targeting PHP 8.2 (with bookworm) post-upgrade, and then whatever PHP version trixie gives us when that is eventually released (I'm assuming it'll probably be at least 8.4).

At the moment, I'm not hugely sure how to run the upgrade as this will be the first major version bump of PHP in quite a long time. My usual preference with larger/riskier upgrades is to break them down into smaller bits that can be applied a little bit at a time (where possible) to reduce the risk. Smarty 5.x should run on 7.4, so I'd definitely prefer if we split that out into a separate change to go live first. The rest I think is probably all interdependent and has to go at once?

includes/Fragments/TemplateOutput.php
```diff
@@ -42,7 +42,7 @@ final protected function assign($name, $value)
      */
     final protected function setUpSmarty()
     {
-        $this->smarty = new Smarty();
+        $this->smarty = new Smarty\Smarty();
```
stwalkerster, line 45: Can we include the class in a `use` statement instead of qualifying it here?

mdaniels5757: Sure, done in #973.

mdaniels5757:
> Smarty 5.x should run on 7.4, so I'd definitely prefer if we split that out into a separate change to go live first. 

Done; see #973.

> The rest I think is probably all interdependent and has to go at once?

Yes.

PR #973, opened 2024-08-19
upgrade smarty to v5.4.0

mdaniels5757:
Breaking this out per #969.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
     "minimum-stability": "stable",
     "require": {
         "php": ">=7.4",
-        "smarty/smarty": "^v4.3.1",
+        "smarty/smarty": "^v5.4.0",
         "twitter/typeahead.js": "^0.11.1",
         "fortawesome/font-awesome": "~6.6.0",
         "spomky-labs/otphp": "^10.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "ff05dad4c2048b7ddd077f2f771b1394",
+    "content-hash": "ca4894afbf91c2c1134f949512736bb1",
     "packages": [
         {
             "name": "bacon/bacon-qr-code",
@@ -54,6 +54,10 @@
             ],
             "description": "BaconQrCode is a QR code generator for PHP.",
             "homepage": "https://github.com/Bacon/BaconQrCode",
+            "support": {
+                "issues": "https://github.com/Bacon/BaconQrCode/issues",
+                "source": "https://github.com/Bacon/BaconQrCode/tree/2.0.8"
+            },
             "time": "2022-12-07T17:46:57+00:00"
         },
         {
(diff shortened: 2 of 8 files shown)


stwalkerster:
I've not spent too much time testing this properly yet, but I am seeing a few things.

It looks like the fix we did for the `implode` modifier in #965  hasn't worked as Smarty [still has an implode modifier][implode] built-in (and throwing deprecation warnings). I've bodged it on team/user.tpl locally, but I'm not sure if this is all cases we need to fix:

```diff
-  <li>{'</li><li>'|implode:$infoContent}</li>
+  <li>{$infoContent|join:'</li><li>'}</li>
```

There's also a lot of deprecation warnings being thrown by `TemplateOutput::setUpSmarty`, as the call to `addPluginsDir` is [also deprecated][addplugins]. The number of deprecation warnings this is throwing on every single page load isn't great. I *think* something like this is working, but I don't know for sure yet. There's probably a much better way to do it than this:

```php
        $pluginsDir = $this->getSiteConfiguration()->getFilePath() . '/smarty-plugins';
//        $this->smarty->addPluginsDir($pluginsDir);

        /** @var DirectoryIterator $file */
        foreach (new DirectoryIterator($pluginsDir) as $file) {
            if ($file->isDot()) {
                continue;
            }

            if ($file->getExtension() != 'php') {
                continue;
            }

            require_once $file->getPathname();

            list($type, $name) = explode('.', $file->getBasename('.php'), 2);

            switch ($type) {
                case 'modifier':
                    $this->smarty->registerPlugin(Smarty::PLUGIN_MODIFIER, $name, 'smarty_modifier_' . $name);
                    break;
                case 'function':
                    $this->smarty->registerPlugin(Smarty::PLUGIN_FUNCTION, $name, 'smarty_function_' . $name);
                    break;
            }
        }
```

I've also looked at the ban list on `/internal.php/bans`, and that's actually properly broken too.

```
Exception 0: Smarty\CompilerException

Syntax error in template "file:bans/list.tpl" on line 39 "{if $ban->getDuration() < time() && $ban->getDuration() !== null}<span class="badge badge-warning">Expired</span>{/if}" unknown modifier 'time'

#0 ./vendor/smarty/smarty/src/Compile/ModifierCompiler.php(74): Smarty\Compiler\Template->trigger_template_error('unknown modifie...', 39, true)
#1 ./vendor/smarty/smarty/src/Compiler/Template.php(498): Smarty\Compile\ModifierCompiler->compile(Array, Object(Smarty\Compiler\Template), Array)
#2 ./vendor/smarty/smarty/src/Compiler/Template.php(1383): Smarty\Compiler\Template->compileModifier(Array, NULL)
#3 ./vendor/smarty/smarty/src/Parser/TemplateParser.php(2726): Smarty\Compiler\Template->compileModifierInExpression('time', Array)
#4 ./vendor/smarty/smarty/src/Parser/TemplateParser.php(2902): Smarty\Parser\TemplateParser->yy_r149()
#5 ./vendor/smarty/smarty/src/Parser/TemplateParser.php(2997): Smarty\Parser\TemplateParser->yy_reduce(149)
#6 ./vendor/smarty/smarty/src/Compiler/Template.php(1269): Smarty\Parser\TemplateParser->doParse(53, ' && ')
#7 ./vendor/smarty/smarty/src/Compiler/Template.php(437): Smarty\Compiler\Template->doCompile('
```

I've not spent a decent amount of time testing this yet, so there may be more things to be uncovered.

[implode]:https://github.com/smarty-php/smarty/blob/master/src/Extension/DefaultExtension.php#L562
[addplugins]:https://github.com/smarty-php/smarty/blob/master/src/Smarty.php#L808

stwalkerster:
I've done a few fixes here myself, mostly based on my comment above. I think I've covered most of the core cases and not found any additional breakage. I'm still mighty nervous about this upgrade, since it's affecting the core of all our display logic.

There's a *lot* of deprecation notices in Smarty 4.5.5 that seem to have been reverted in 5.5.0, specifically around use of static constants from unregistered classes in templates. I'm thankful this is no longer throwing deprecation messages, and the relevant functionality does still seem to be present. I hope use of constants in templates is here to stay.

There's also one vaguely unrelated change to RoleConfigurationBase which suppresses a truckload (read, about 50 per page load) of warnings. In that case it was interpreting one of the role configuration flags as an actual page for permission assignments. We've already got a blocklist of array keys that aren't pages, so I added the errant the flag to that list. It should have been done as part of the introduction of global roles, but it was missed.

composer.json
```diff
@@ -3,7 +3,7 @@
     "minimum-stability": "stable",
     "require": {
         "php": ">=7.4",
-        "smarty/smarty": "^v4.3.1",
+        "smarty/smarty": "^v5.4.0",
```
FastLizard4, line 6: It looks like we're actually explicitly targeting v5.5.0 based on the discussion in the PR; should the title be adjusted and this changed to 5.5.0?

FastLizard4 (commented):
One question, otherwise LGTM (with an optional nit)

composer.json
stwalkerster: The PR title isn't particularly important; the `composer.json` file probably also isn't because any install will use the locked version v5.5.0 from `composer.lock`. 

This version constraint permits any 5.* version >= 5.4.0, so it could be argued that since we depend on 5.5.0 specifically for the reverted deprecations we should specify that version. Realistically though, our locked version is always going to be above the specified minimum version in composer.json simply because of Dependabot. I'm minded to change this anyway, but it feels like a nitpick too. What do you think?

PR #989, opened 2024-11-04
Fix unit tests

stwalkerster:
One test case was changed per https://meta.wikimedia.org/w/index.php?title=Access_to_nonpublic_personal_data_policy/Noticeboard&diff=prev&oldid=27702669


FastLizard4 (approved):
image

PR #1003, opened 2024-12-27
Add ability to appeal suspensions from within the tool

stwalkerster:
This change does several things:
* Merges the concepts of suspended and declined users into a single "deactivated" user
* Adds a form to appeal deactivations; submission of which will move a user into the "New" state and write the message to the log
* Adds the ability to disable the ability for a user to appeal in case of abuse
* Hides the comments from account status changes (specifically suspension/appeals) from all but tool admins and the affected user themselves


FastLizard4 (approved):
Code and some local testing both look good to me, though I noted one non-blocking low-priority issue that can be handled in a follow-up PR: #1013

PR #1019, opened 2025-03-29
Allow UA to be NULL

### LINKED ISSUE #881: Make `request.useragent` column in database nullable
Currently, the `request.useragent` column in the database is defined as `TEXT NOT NULL`, which means that any requests we receive where we don't receive a `User-Agent` HTTP header will cause an exception and fail. Currently we only see this with requests that appear to be fuzzed for automated scanning or similar, but it's still not an ideal behavior.

To fix, we should make the `request.useragent` column nullable.

methecooldude:
Fixes #881

### DIFF
--- a/sql/patches/patch51-useragent-nullable.sql
+++ b/sql/patches/patch51-useragent-nullable.sql
@@ -1,20 +1,3 @@
--- -----------------------------------------------------------------------------
--- Hey!
--- 
--- This is a new patch-creation script which SHOULD stop double-patching and
--- running patches out-of-order.
---
--- If you're running patches, please close this file, and run this from the 
--- command line:
---   $ mysql -u USERNAME -p SCHEMA < patchXX-this-file.sql
--- where:
---      USERNAME = a user with CREATE/ALTER access to the schema
---      SCHEMA = the schema to run the changes against
---      patch-XX-this-file.sql = this file
---
--- If you are writing patches, you need to copy this template to a numbered 
--- patch file, update the patchversion variable, and add the SQL code to upgrade
--- the database where indicated below.
 /******************************************************************************
  * Wikipedia Account Creation Assistance tool                                 *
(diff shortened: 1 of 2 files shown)


sql/patches/patch51-useragent-nullable.sql
```diff
@@ -0,0 +1,70 @@
...
+-- command line:
+--   $ mysql -u USERNAME -p SCHEMA < patchXX-this-file.sql
+-- where:
+--      USERNAME = a user with CREATE/ALTER access to the schema
+--      SCHEMA = the schema to run the changes against
+--      patch-XX-this-file.sql = this file
+--
+-- If you are writing patches, you need to copy this template to a numbered 
+-- patch file, update the patchversion variable, and add the SQL code to upgrade
+-- the database where indicated below.
```
stwalkerster, line 17: nitpick: This header comment can probably be deleted

sql/patches/patch51-useragent-nullable.sql
```diff
@@ -0,0 +1,70 @@
...
+ * This is free and unencumbered software released into the public domain.    *
+ * Please see LICENSE.md for the full licencing statement.                    *
+ ******************************************************************************/
+DROP PROCEDURE IF EXISTS SCHEMA_UPGRADE_SCRIPT;
+DELIMITER ';;'
+CREATE PROCEDURE SCHEMA_UPGRADE_SCRIPT() BEGIN
+    -- -------------------------------------------------------------------------
+    -- Developers - set the number of the schema patch here!
+    -- -------------------------------------------------------------------------
+    DECLARE patchversion INT DEFAULT 51;
```
stwalkerster, line 31: You'll need to change SiteConfiguration too as that has a field with the latest patch number in it

stwalkerster (changes requested):

stwalkerster (approved):

PR #1020, opened 2025-03-29
Retry if Nonce Error

### LINKED ISSUE #974: Auto-creation error: Nonce already used
Not sure if this is on our end (possibly due to rapid account creations) or MediaWiki's, but I just got this error in production (request #346825):

```mwoauth-invalid-authorization: The authorization headers in your request are not valid: Nonce already used: [nonce redacted by me]```

I don't recall getting this one before, but some query like `SELECT * FROM jobqueue WHERE status = 'failed' AND error LIKE 'mwoauth-invalid-authorization%';` should make clear whether this is a one-off or not.

It looks like Pywikibot retries when this happens (https://gerrit.wikimedia.org/r/c/pywikibot/core/+/289582), would doing that in our code be desirable? If so, I can make a PR.

### PR DESCRIPTION
**Pull request by methecooldude**
Fixes #974
Probably not the best way of doing this... feedback welcome :P

### CODE REVIEW

File: includes/ConsoleTasks/RunJobQueueTask.php [hunk 1]
```diff
@@ -106,13 +106,19 @@ public function execute()
...
                     $database->beginTransaction();
-
+                    
                     /** @var JobQueue $job */
                     $job = JobQueue::getById($job->getId(), $database);
                     $job->setDatabase($database);
-                    $job->setStatus(JobQueue::STATUS_FAILED);
-                    $job->setError($ex->getMessage());
-                    $job->setAcknowledged(0);
+                    if (str_contains($ex->getMessage, 'mwoauth-invalid-authorization')) {
```
stwalkerster, line 113: This string appears to only come from [this exception handler](https://gerrit.wikimedia.org/g/mediawiki/extensions/OAuth/+/7b4988f1613f958938fa48f074fec546651ace2a/src/SessionProvider.php#151), but it's also a substring used by a [number of different errors](https://codesearch.wmcloud.org/search/?q=mwoauth-invalid-authorization&files=&excludeFiles=i18n&repos=Extension%3AOAuth).

I think this will retry any authorization problem, including quite a few scenarios where a retry won't help at all.

The `Nonce already used` error [appears to be hard-coded at the moment](https://gerrit.wikimedia.org/g/mediawiki/extensions/OAuth/+/7b4988f1613f958938fa48f074fec546651ace2a/src/Lib/OAuthServer.php#267) so it might be better to match on that string instead?

stwalkerster: I'm also concerned that this might cause a job to loop indefinitely through the job queue. I was wondering if we could use the acknowledged flag to determine if a job has been retried already, but it looks like we set it to null when transitioning to "waiting": https://github.com/enwikipedia-acc/waca/blob/997d771885c5f63aa86c209de9e661cc7a264ad8/includes/ConsoleTasks/RunJobQueueTask.php#L56-L58

I think if we can filter on just the nonce error, the risk of a loop is mitigated. I'd hope as a community we'd also notice if jobs got stuck in a loop.

includes/ConsoleTasks/RunJobQueueTask.php
```diff
@@ -106,13 +106,19 @@ public function execute()
...
+                        $job->setError('Receieved error with authorization headers. Will retry');
+                        $job->setStatus(JobQueue::STATUS_READY);
+                    } else {
+                        $job->setStatus(JobQueue::STATUS_FAILED);
+                        $job->setError($ex->getMessage());
+                        $job->setAcknowledged(0);
+                    }
                     $job->save();
 
                     Logger::backgroundJobIssue($this->getDatabase(), $job);
```
stwalkerster, line 124: I think it'd be good to generate a requeue log entry as well to make it more clear what's going on.

stwalkerster (changes requested):
Truly sorry about the time it's taken me to get to reviewing this; and sorry it's a change request too.

PR #1028, opened 2025-05-09
IPv6 Fixes

### LINKED ISSUE #1027: Location lookups don't work for IPv6
Reported by @methecooldude on IRC

methecooldude:
Fix #1027 
Updated bans lookup to support IPV6
Change how rDNS is retrieved
Add the v6 private ranges and remove the filter flag so it can make API calls to iplocation.io on a v6 address
Refactor how IP ranges are worked out

### DIFF
--- a/includes/Helpers/BanHelper.php
+++ b/includes/Helpers/BanHelper.php
@@ -173,18 +173,17 @@ SQL;
 SELECT b.* FROM ban b
 LEFT JOIN netmask n ON 1 = 1
     AND n.cidr = b.ipmask
-    AND n.protocol = CASE LENGTH(b.ip) WHEN 4 THEN 4 WHEN 16 THEN 6 END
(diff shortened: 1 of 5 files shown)


includes/Helpers/BanHelper.php
```diff
@@ -173,18 +173,17 @@ private function getBansForRequestFromDatabase(Request $request): array
 SELECT b.* FROM ban b
 LEFT JOIN netmask n ON 1 = 1
     AND n.cidr = b.ipmask
-    AND n.protocol = CASE LENGTH(b.ip) WHEN 4 THEN 4 WHEN 16 THEN 6 END
+    AND n.protocol = CASE WHEN INET6_ATON(b.ip) IS NOT NULL THEN 6 ELSE 4 END
```
stwalkerster, line 176: It looks like [`inet6_aton`][1] returns a non-null value for IPv4 addresses too?


[1]: https://mariadb.com/kb/en/inet6_aton/

PR #1029, opened 2025-05-10
Add configurable global deny role

stwalkerster:
This adds a configurable role definition which is applied to every ACL check by everyone, logged in or not. This role definition is filtered to only permit `deny` actions so additional access cannot be granted via the configuration file.

The intent of this is to act as an emergency switch-off for any functionality in the tool by denying everyone access to it.

### DIFF
--- a/includes/Security/RoleConfigurationBase.php
+++ b/includes/Security/RoleConfigurationBase.php
@@ -18,11 +18,14 @@ abstract class RoleConfigurationBase
 
     protected array $roleConfig;
     protected array $identificationExempt;
+    private array $globallyDenied;
 
-    protected function __construct(array $roleConfig, array $identificationExempt)
+    protected function __construct(array $roleConfig, array $identificationExempt, array $globallyDenied)
     {
         $this->roleConfig = $roleConfig;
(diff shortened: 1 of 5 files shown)


includes/Security/RoleConfigurationBase.php
```diff
@@ -18,11 +18,14 @@ abstract class RoleConfigurationBase
...
     protected array $identificationExempt;
+    private array $globallyDenied;
 
-    protected function __construct(array $roleConfig, array $identificationExempt)
+    protected function __construct(array $roleConfig, array $identificationExempt, array $globallyDenied)
     {
         $this->roleConfig = $roleConfig;
         $this->identificationExempt = $identificationExempt;
+
+        $this->globallyDenied = self::constructDenyOnlyRole($globallyDenied);;
```
FastLizard4, line 28: nit: Extra semicolon

FastLizard4 (approved):
Just one nit, otherwise LGTM

includes/Security/RoleConfigurationBase.php
FastLizard4: So uh this is still here lol

PR #1030, opened 2025-05-12
Clean up breakage in some tools caused by IPv6

### LINKED ISSUE #1026: Some IP tools don't work with IPv6
Specifically StopForumSpam and IPalyzer fail due to the urlencoding.

I've already done some digging into this, and I think we can skip some of the urlencoding and instead lean on filter_var in redir.php.

stwalkerster:
Fixes #1026

### DIFF
--- a/smarty-plugins/modifier.cidr.php
+++ b/smarty-plugins/modifier.cidr.php
@@ -1,0 +1,64 @@
+<?php
+/******************************************************************************
+ * Wikipedia Account Creation Assistance tool                                 *
+ * ACC Development Team. Please see team.json for a list of contributors.     *
+ *                                                                            *
+ * This is free and unencumbered software released into the public domain.    *
+ * Please see LICENSE.md for the full licencing statement.                    *
+ ******************************************************************************/
+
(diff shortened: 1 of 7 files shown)


smarty-plugins/modifier.cidr.php
```diff
@@ -0,0 +1,64 @@
...
+            if ($prefix < 8) {
+                $bits = $prefix;
+            }
+            $mask = $bits === 0 ? 0 : (0xFF << (8 - $bits)) & 0xFF;
+            $bin .= chr(ord($ipBin[$i]) & $mask);
+            $prefix -= $bits;
+            if ($prefix <= 0) {
+                $prefix = 0;
+            }
+        }
```
FastLizard4, line 58: 🤮

FastLizard4 (approved):

smarty-plugins/modifier.cidr.php
stwalkerster: ... this is precisely why this function is covered by copious amounts of tests...

PR #1037, opened 2025-05-22
Docker - Add Defaults

methecooldude:
OAuth is a PITA... disable it by default on Docker builds


stwalkerster (dismissed):
LGTM, but will hold of merging for now

stwalkerster (approved):

PR #1058, opened 2025-09-06
Change where spur points to

dqwiki:
image


stwalkerster:
Yes, the Spur link has been broken for a little while; the correct link would be https://app.spur.us/search?q=1.2.3.4 for authenticated users.

I'm not sure if I'd prefer we had an authenticated link or not for those of us who do have accounts on Spur; it's certainly irksome that auth vs unauth is two different URLs. Thoughts?

stwalkerster (approved):
Approving since this is strictly an improvement on the current situation, even if I'm not sure this is the ideal end goal.

PR #1093, opened 2026-06-06
Handle TBL API being dead

methecooldude:
Turns out that if the API fails for the TBL check, everything ends up in Flagged Users.

Added check to BlacklistHelper to throw an Exception if there is no result

Added check to PageViewRequest to check the data isn't an empty string (in addition to the check for false) before as it seems that's the issue why requests are ending up in Flagged Users.

Also, add a check to the if statement just in case a default creation mode has not been set
Fixes #1081  (hopefully) (Yes, Rich does not know how to Git properly...)

### DIFF
--- a/includes/Helpers/BlacklistHelper.php
+++ b/includes/Helpers/BlacklistHelper.php
@@ -103,7 +103,7 @@ class BlacklistHelper implements IBlacklistHelper
 
         $parameters = array(
             'action'       => 'titleblacklist',
-            'format'       => 'php',
+            'format'       => 'json',
             'tbtitle'      => $username,
             'tbaction'     => 'new-account',
             'tbnooverride' => true,
@@ -111,7 +111,7 @@ class BlacklistHelper implements IBlacklistHelper
 
         $apiResult = $this->httpHelper->get($endpoint, $parameters);
 
-        $data = unserialize($apiResult);
+        $data = json_decode($apiResult, true);
 
         return $data['titleblacklist'];
     }
--- a/includes/Helpers/BlacklistHelper.php
+++ b/includes/Helpers/BlacklistHelper.php
@@ -63,14 +63,6 @@ class BlacklistHelper implements IBlacklistHelper
             ExceptionHandler::logExceptionToDisk($ex, $this->siteConfiguration);
             return false;
         }
-        
-        if (!isset($result['result'])) {
-            if (isset($result['error']['info'])) {
-                throw new Exception("Unrecognised API response to blacklist query: " . $result['error']['info']);
-            }
-            
-            throw new Exception("Unrecognised API response to blacklist query.");
-        }
 
         if ($result['result'] === 'ok') {
             // not blacklisted
--- a/includes/Helpers/BlacklistHelper.php
+++ b/includes/Helpers/BlacklistHelper.php
@@ -63,6 +63,14 @@ class BlacklistHelper implements IBlacklistHelper
             ExceptionHandler::logExceptionToDisk($ex, $this->siteConfiguration);
             return false;
         }
+        
+        if (!isset($result['result'])) {
+            if (isset($result['error']['info'])) {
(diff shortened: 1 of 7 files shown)


stwalkerster:
OK, so I think there's more going on here than meets the eye.

There seems to be two calls to the title blacklist; one in PageViewRequest via `$this->getBlacklistHelper()->isBlacklisted($request->getName());` which is called to evaluate the request against the blacklist when looking at the request on the interface, and one in RequestValidationHelper L330ish which is called when the request is first saved and this is what should determine if a request is sent to the flagged users queue or not. The behaviour we're seeing (deferrals logged at request creation time) points to RequestValidationHelper being the issue causing the deferrals.

However, we're *also* getting AntiSpoof exceptions being logged in the tool:

```
Unrecognised API response to query.

#0 ./includes/Validation/RequestValidationHelper.php(296): Waca\Providers\CachedApiAntispoofProvider->getSpoofs()
#1 ./includes/Validation/RequestValidationHelper.php(224): Waca\Validation\RequestValidationHelper->checkAntiSpoof()
#2 ./includes/Pages/Request/PageRequestAccount.php(301): Waca\Validation\RequestValidationHelper->postSaveValidations()
#3 ./includes/Pages/Request/PageRequestAccount.php(88): Waca\Pages\Request\PageRequestAccount->handleFormPost()
#4 ./includes/Tasks/PageBase.php(108): Waca\Pages\Request\PageRequestAccount->dynamic()
#5 ./includes/Tasks/PageBase.php(368): Waca\Tasks\PageBase->runPage()
#6 ./includes/Tasks/PublicInterfacePageBase.php(24): Waca\Tasks\PageBase->execute()
#7 ./includes/WebStart.php(193): Waca\Tasks\PublicInterfacePageBase->execute()
#8 ./includes/WebStart.php(112): Waca\WebStart->main()
#9 ./index.php(28): Waca\WebStart->run()
#10 {main}
```

stwalkerster:
`Unrecognized value for parameter "format": php.` is what the API is returning.

I think most of our API calls use the PHP format, and it looks like it's recently been removed. We'll need to migrate all our API calls to a different format.

methecooldude:
> `Unrecognized value for parameter "format": php.` is what the API is returning.
> 
> I think most of our API calls use the PHP format, and it looks like it's recently been removed. We'll need to migrate all our API calls to a different format.

Oh, great...

includes/Helpers/BlacklistHelper.php
```diff
@@ -63,6 +63,14 @@ public function isBlacklisted($username)
...
             return false;
         }
+        
+        if (!isset($result['result'])) {
+            if (isset($result['error']['info'])) {
+                throw new Exception("Unrecognised API response to blacklist query: " . $result['error']['info']);
+            }
+            
+            throw new Exception("Unrecognised API response to blacklist query.");
+        }
```
stwalkerster: I'm not certain this is a good idea. If these exceptions aren't caught by anything (and I can't see anything that *would* catch them), this'll end up with an "Oops something went wrong, have this error ID" white screen of death.

includes/Providers/IpLocationProvider.php
```diff
@@ -91,13 +90,17 @@ private function getResult($ip)
     {
         try {
             if (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4)) {
-                $xml = $this->httpHelper->get($this->getApiBase(), array(
+                $json = $this->httpHelper->get($this->getApiBase(), array(
                     'key'    => $this->apiKey,
                     'ip'     => $ip,
-                    'format' => 'xml',
+                    'format' => 'json',
```
stwalkerster: This is definitely a later problem - but this is currently serialize()'d into the database. IMHO we should be storing it in JSON there as well. That'd either need a data migration for converting the existing data or just binning the existing data.

stwalkerster (changes requested):
LGTM apart from that exception being thrown

stwalkerster (approved):

